PR #1, opened 2020-06-09
chore: Initial project setup

atiqueahmedziad:
The following are added: 
- eslint
- prettier
- babel (for supporting es6 module in jest)
- jest
- travis CI & coveralls


package.json
```diff
@@ -0,0 +1,43 @@
...
+  "keywords": [
+    "activityLog",
+    "extension activity monitor",
+    "firefox extension activity monitor",
+    "firefox",
+    "firefox extension logs",
+    "extension acitivty logs",
+    "mozilla"
+  ],
+  "author": "Mozilla <mozilla@mozilla.org>",
```
atiqueahmedziad, line 27: I am not sure about it. Need your suggestion.

.travis.yml
```diff
@@ -0,0 +1,12 @@
...
+
+node_js:
+  - '14'
+
+script:
+  - npm run prettier:check
+  - npm run lint:check
+  - npm run test:coveralls
+
+notifications:
```
atiqueahmedziad, line 11: How do we setup notification. Need suggestion here.

package.json
rpl: I think we could set the author property of the package.json file to "Mozilla Add-ons Team", [as in the addons-linter repo](https://github.com/mozilla/addons-linter/blob/713cddc11ce64f9177040acb5f77f39176cb07ee/package.json#L35).

We could also use the same string for the author property in the manifest.json file.

package.json
```diff
@@ -0,0 +1,43 @@
...
+    "type": "git",
+    "url": "git+https://github.com/mozilla/extension-activity-monitor.git"
+  },
+  "keywords": [
+    "activityLog",
+    "extension activity monitor",
+    "firefox extension activity monitor",
+    "firefox",
+    "firefox extension logs",
+    "extension acitivty logs",
```
rpl, line 24: typo: acitivty => activity

.travis.yml
rpl: For the addons-linter the travis notifications are being sent to slack, on web-ext the travis notification are still configured to go to irc (which we don't use anymore, and so it is kind of /dev/null :-P).

Personally I think that we can leave the travis notification disabled (at least for now).

atiqueahmedziad:
> Follows an initial round of review comments.
> 
> Besides the inline review comments, I propose to keep using a convention for the commit messages (even if this repository isn't yet using any automation to enforce that), the first commit is already using that convention (the commit message is "chore: ..." to recognize it as a change that doesn't need to be listed in a changelog).
> 
> See https://www.conventionalcommits.org/en/v1.0.0/ for more information about this convention (and you may look to the web-ext repo as a project that is actively using these kind of convention to generate a changelog in a semi-automated way).
> 
> In my opinion this would also be a "chore" commit and so the commit message could be "chore: Initial project setup".

Thanks for the feedbacks.
For now, I have updated the pr with new commit, so you can view the new changes that I made. Once we are done with changes, I will squash commits and update the commit message to "chore: Initial project setup".

rpl:
@atiqueahmedziad I enabled travis on this repo and so the travis config did start to run, would you mind to look into making the travis job to pass? 
([from a quick look to the logs on travis it seems that it is failing just because jest is complaining about the test file being completely empty](https://travis-ci.org/github/mozilla/extension-activity-monitor/jobs/696463744#L230-L233), which we may fix by just adding a "placeholder" test case, then we can replace it with some real test cases in the next pull request).

atiqueahmedziad:
@rpl I have added a placeholder test. It seems [coveralls hasn't been activated for this repo](https://travis-ci.org/github/mozilla/extension-activity-monitor/builds/696562981#L242) but I have that running in [`.travis.yml`](https://github.com/mozilla/extension-activity-monitor/pull/1/files#diff-354f30a63fb0907d4ad57269548329e3R9) file. Would you activate coveralls for this repository ? I think I can do that too if you permit me.

src/background.js
```diff
@@ -0,0 +1,5 @@
+const dummy = {
+  name: 'john doe',
+};
+
+export default dummy;
```
rpl, line 5: To use ES6 modules in a background page we will have to use `background.page` instead of `background.scripts` in the manifest.json file and define a background HTML page and then load the scripts into it as ES6 modules [using script tags with type "module"].

rpl:
> @rpl I have added a placeholder test and updated the commit title. It seems [coveralls hasn't been activated for this repo](https://travis-ci.org/github/mozilla/extension-activity-monitor/builds/696562981#L242) but I have that running in [`.travis.yml`](https://github.com/mozilla/extension-activity-monitor/pull/1/files#diff-354f30a63fb0907d4ad57269548329e3R9) file. Would you activate coveralls for this repository ? I think I can do that too if you permit me.

I just activated this repository on coveralls and forced travis to re-run the build (which is now green).

I added another review comment related to the background.js file being written as an ES6 module, which isn't going to work while using `background.scripts` in the manifest.

Let's deal with that before merging this pull request.

atiqueahmedziad:
> I added another review comment related to the background.js file being written as an ES6 module, which isn't going to work while using `background.scripts` in the manifest.

Sorry, didn't notice that. Just to make it a little less complicated, I have moved from `background.scripts` and added a separate module and tested on it. :-)

PR #2, opened 2020-06-10
feat: save and load module added

atiqueahmedziad:
The functionalities to load logs file (json format) and save log file are added.

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -1,6 +1,7 @@
 {
   "extends": ["eslint:recommended", "plugin:import/errors", "prettier"],
   "plugins": ["import"],
+  "parser": "babel-eslint",
   "parserOptions": {
     "ecmaVersion": 2019,
     "sourceType": "module"
--- a/package-lock.json
+++ b/package-lock.json
@@ -1430,6 +1430,76 @@
         "@types/istanbul-lib-report": "*"
       }
     },
+    "@types/jest": {
+      "version": "26.0.0",
+      "resolved": "https://registry.npmjs.org/@types/jest/-/jest-26.0.0.tgz",
+      "integrity": "sha512-/yeMsH9HQ1RLORlXAwoLXe8S98xxvhNtUz3yrgrwbaxYjT+6SFPZZRksmRKRA6L5vsUtSHeN71viDOTTyYAD+g==",
+      "dev": true,
+      "requires": {
+        "jest-diff": "^25.2.1",
+        "pretty-format": "^25.2.1"
+      },
+      "dependencies": {
+        "@jest/types": {
+          "version": "25.5.0",
+          "resolved": "https://registry.npmjs.org/@jest/types/-/types-25.5.0.tgz",
+          "integrity": "sha512-OXD0RgQ86Tu3MazKo8bnrkDRaDXXMGUqd+kTtLtK1Zb7CRzQcaSRPPPV37SvYTdevXEBVxe0HXylEjs8ibkmCw==",
+          "dev": true,
+          "requires": {
+            "@types/istanbul-lib-coverage": "^2.0.0",
+            "@types/istanbul-reports": "^1.1.1",
+            "@types/yargs": "^15.0.0",
+            "chalk": "^3.0.0"
+          }
+        },
+        "chalk": {
+          "version": "3.0.0",
+          "resolved": "https://registry.npmjs.org/chalk/-/chalk-3.0.0.tgz",
+          "integrity": "sha512-4D3B6Wf41KOYRFdszmDqMCGq5VV/uMAB273JILmO+3jAlh8X4qDtdtgCR3fxtbLEMzSx22QdhnDcJvu2u1fVwg==",
+          "dev": true,
+          "requires": {
+            "ansi-styles": "^4.1.0",
+            "supports-color": "^7.1.0"
+          }
+        },
+        "diff-sequences": {
+          "version": "25.2.6",
+          "resolved": "https://registry.npmjs.org/diff-sequences/-/diff-sequences-25.2.6.tgz",
+          "integrity": "sha512-Hq8o7+6GaZeoFjtpgvRBUknSXNeJiCx7V9Fr94ZMljNiCr9n9L8H8aJqgWOQiDDGdyn29fRNcDdRVJ5fdyihfg==",
+          "dev": true
+        },
+        "jest-diff": {
+          "version": "25.5.0",
+          "resolved": "https://registry.npmjs.org/jest-diff/-/jest-diff-25.5.0.tgz",
+          "integrity": "sha512-z1kygetuPiREYdNIumRpAHY6RXiGmp70YHptjdaxTWGmA085W3iCnXNx0DhflK3vwrKmrRWyY1wUpkPMVxMK7A==",
+          "dev": true,
+          "requires": {
+            "chalk": "^3.0.0",
+            "diff-sequences": "^25.2.6",
+            "jest-get-type": "^25.2.6",
+            "pretty-format": "^25.5.0"
+          }
+        },
+        "jest-get-type": {
+          "version": "25.2.6",
+          "resolved": "https://registry.npmjs.org/jest-get-type/-/jest-get-type-25.2.6.tgz",
+          "integrity": "sha512-DxjtyzOHjObRM+sM1knti6or+eOgcGU4xVSb2HNP1TqO4ahsT+rqZg+nyqHWJSvWgKC5cG3QjGFBqxLghiF/Ig==",
+          "dev": true
+        },
+        "pretty-format": {
+          "version": "25.5.0",
+          "resolved": "https://registry.npmjs.org/pretty-format/-/pretty-format-25.5.0.tgz",
+          "integrity": "sha512-kbo/kq2LQ/A/is0PQwsEHM7Ca6//bGPPvU6UnsdDRSKTWxT/ru/xb88v4BJf6a69H+uTytOEsTusT9ksd/1iWQ==",
+          "dev": true,
+          "requires": {
+            "@jest/types": "^25.5.0",
+            "ansi-regex": "^5.0.0",
+            "ansi-styles": "^4.0.0",
+            "react-is": "^16.12.0"
+          }
+        }
+      }
+    },
     "@types/json5": {
       "version": "0.0.29",
       "resolved": "https://registry.npmjs.org/@types/json5/-/json5-0.0.29.tgz",
@@ -1668,6 +1738,20 @@
       "resolved": "https://registry.npmjs.org/aws4/-/aws4-1.10.0.tgz",
       "integrity": "sha512-3YDiu347mtVtjpyV3u5kVqQLP242c06zwDOgpeRnybmXlYYsLbtTrUBUm8i8srONt+FWobl5aibnU1030PeeuA==",
       "dev": true
+    },
+    "babel-eslint": {
+      "version": "10.1.0",
+      "resolved": "https://registry.npmjs.org/babel-eslint/-/babel-eslint-10.1.0.tgz",
+      "integrity": "sha512-ifWaTHQ0ce+448CYop8AdrQiBsGrnC+bMgfyKFdi6EsPLTAWG+QfyDeM6OH+FmWnKvEq5NnBMLvlBUPKQZoDSg==",
+      "dev": true,
+      "requires": {
+        "@babel/code-frame": "^7.0.0",
+        "@babel/parser": "^7.7.0",
+        "@babel/traverse": "^7.7.0",
+        "@babel/types": "^7.7.0",
+        "eslint-visitor-keys": "^1.0.0",
+        "resolve": "^1.12.0"
+      }
     },
     "babel-jest": {
       "version": "26.0.1",
@@ -2352,22 +2436,22 @@
       }
     },
     "es-abstract": {
-      "version": "1.17.5",
-      "resolved": "https://registry.npmjs.org/es-abstract/-/es-abstract-1.17.5.tgz",
-      "integrity": "sha512-BR9auzDbySxOcfog0tLECW8l28eRGpDpU3Dm3Hp4q/N+VtLTmyj4EUN088XZWQDW/hzj6sYRDXeOFsaAODKvpg==",
+      "version": "1.17.6",
+      "resolved": "https://registry.npmjs.org/es-abstract/-/es-abstract-1.17.6.tgz",
+      "integrity": "sha512-Fr89bON3WFyUi5EvAeI48QTWX0AyekGgLA8H+c+7fbfCkJwRWRMLd8CQedNEyJuoYYhmtEqY92pgte1FAhBlhw==",
       "dev": true,
       "requires": {
         "es-to-primitive": "^1.2.1",
         "function-bind": "^1.1.1",
         "has": "^1.0.3",
         "has-symbols": "^1.0.1",
-        "is-callable": "^1.1.5",
-        "is-regex": "^1.0.5",
+        "is-callable": "^1.2.0",
+        "is-regex": "^1.1.0",
         "object-inspect": "^1.7.0",
         "object-keys": "^1.1.1",
         "object.assign": "^4.1.0",
-        "string.prototype.trimleft": "^2.1.1",
-        "string.prototype.trimright": "^2.1.1"
+        "string.prototype.trimend": "^1.0.1",
+        "string.prototype.trimstart": "^1.0.1"
       }
     },
     "es-to-primitive": {
@@ -2551,9 +2635,9 @@
       }
     },
     "eslint-import-resolver-node": {
-      "version": "0.3.3",
-      "resolved": "https://registry.npmjs.org/eslint-import-resolver-node/-/eslint-import-resolver-node-0.3.3.tgz",
-      "integrity": "sha512-b8crLDo0M5RSe5YG8Pu2DYBj71tSB6OvXkfzwbJU2w7y8P4/yo0MyF8jU26IEuEuHF2K5/gcAJE3LhQGqBBbVg==",
+      "version": "0.3.4",
+      "resolved": "https://registry.npmjs.org/eslint-import-resolver-node/-/eslint-import-resolver-node-0.3.4.tgz",
+      "integrity": "sha512-ogtf+5AB/O+nM6DIeBUNr2fuT7ot9Qg/1harBfBtaP13ekEWFQEEMP94BCB7zaNW3gyY+8SHYF00rnqYwXKWOA==",
       "dev": true,
       "requires": {
         "debug": "^2.6.9",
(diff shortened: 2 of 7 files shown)


rpl:
> @rpl do we need test for these ? In case we do it would be helpful if you can give me some suggestion on it. I tried for sometime, but not understanding how do I proceed to save/ load file in the test file.

@atiqueahmedziad follows some ideas and suggestions, a first part is about planning tests (about some reasons and principles that may be useful to guide ourselves while working on tests) and a second part related to what Jest provides to help us to test our modules (and link to the related docs). 

## Planning tests

Tests are important not only because they are going to ensure the integrity of the project while we apply more changes over the time, they are also an opportunity to double-check our designs, asking ourselves questions about it. Testing should help us to look to the code we are writing from different angles.

Just as an example, some "design" questions to ask ourselves may be:

- what are the responsibilities of the Save and Load classes
- how are they supposed to be used by the other components of the project
- what are the other components that they depends on?

And some other questions specifically related to "testing" could be: 

- what kind of scenario would be good to test?
- if some test is hard to write, could changing the design a bit be helpful to make it more easily testable?
- what should I mock to cover the scenario that I'm interested in, what the mock should do?
- what expectations should the test check?

While creating a new project or a new module, we usually start by writing unit tests, right?
While writing the unit tests our goal is testing a particular unit of work in isolation, with the external dependencies of that unit often mocked and/or instrumented.

## Jest, JSDOM and other custom mocks

Some pointers from an "implementation strategies" perspective:

The [default testing environment for Jest is based on "jsdom"](https://jestjs.io/docs/en/configuration#testenvironment-string), and ["jsdom" provides a browser-like environment](https://github.com/jsdom/jsdom). Basically jsdom already mocks some of the APIs available in a webpage.

As an example setTimeout is already mocked, and jest provides two helpers to help us testing code that are using setTimeout, see [Jest Timer Mocks doc page here](https://jestjs.io/docs/en/timer-mocks.html).

Besides setTimeout, the code in these ES6 modules are also using other WebAPI or WebExtensions API that we may need to mock manually. e.g. internally Save does use `URL.createObjectURL`/`revokeObjectURL` and a subset of the `browser.downloads` WebExtensions API.

Jest provides helpers to create mocks (e.g. see the Jest [Manual Mocks](https://jestjs.io/docs/en/manual-mocks) and [Mock Functions](https://jestjs.io/docs/en/mock-function-api) doc pages)  and checks some expectations on them (e.g. see [expect(...).toHaveBeenCalledWith](https://jestjs.io/docs/en/expect#tohavebeencalledwitharg1-arg2-)).

By putting all this together, we can basically test almost everything we want.
Just as an example:

### let's decide what scenario we are going to test first:
 
```js
test('Save saveAsJSON should create a blob url and download it', async () => {
  ...
});
```

### then inside this new test we can start mocking what the unit we are testing depends on:

```js
  // Create window.URL mocks.
  const createObjectURL = jest.fn();
  const revokeObjectURL = jest.fn();
  window.URL = { createObjectURL, revokeObjectURL };

  // Create browser.downloads mocks.
  const download = jest.fn();
  const addListener = jest.fn();
  const removeListener = jest.fn();
  window.browser = {
    downloads: {
      download,
      onChanged: { addListener, removeListener },
    },
  };
```

### then we can mock the implementation

For some of the mocks we may need them to do something special, to be able to trigger the scenario we are going to test, e.g:

- we may want to let URL.createObjectURL to return a fake value as url when it is called:
```js
  createObjectURL.mockReturnValue("fake-blob-url");
```

- and we may want the test to wait for URL.revokeObjectURL to be called, because that is going to happen asynchronously:

```js
  // Mock URL.revokeObjectURL implementation to await
  // for it to be called.
  const promiseRevokedURL = new Promise(resolve => {
    revokeObjectURL.mockImplementation(() => resolve());
  });
```

- If setTimeout is needed (more question on this in a review comment), then we should ask jest to mock it (See Jest [Timer Mocks](https://jestjs.io/docs/en/timer-mocks.html) doc page):

```js
  // Mocks setTimeout.
  jest.useFakeTimers();
```

- To be able to cover the code that should be executed when the downloads.onChanged event is going to be emitted, we will likely need to mock onChanged.addListener a bit:

```js
  // Mock downloads.onChanged.addListener.
  let listener;
  addListener.mockImplementation(
    (cb) => { listener = cb; }
  );
```

- and then mock `browser.downloads.download` to make it also call the listener to mock what would happen for real when the downloads.onChanged API events are going to be emitted:

```js
  // Mock downloads.download.
  download.mockImplementation(() => {
    // Fake incomplete download onChanged event fired.
    listener({state: {...}});

    // Fake complete download onChanged event fired.
    Promise.resolve().then(() => listener({
      state: { current: "complete" },
    })).then(() => {
      // Force run setTimeout callbacks. 
      jest.runAllTimers();
    });
  })
```

- we may also want to be sure that the test is also going to wait that the listener subscribed is going to be removed, and so we may mock removeListener a bit:

```js
  // Mock downloads.onChanged.removeListener.
  const promiseRemovedListener = new Promise(resolve => {
    removeListener.mockImplementation(() => resolve());
  });
```

### At this point we should be ready to load the unit we want to test in the testing environment we have prepared and exercise the module API:

```js
  const Save = require("../src/lib/Save").default;
  const sv1 = new Save([{prop1: "log1"}]);
  sv1.saveAsJSON();

  await Promise.all([promiseRevokedURL, promiseRemovedListener]);
```

### And then check that our expectations are met:

```js
  expect(download).toHaveBeenCalledTimes(1);
  expect(download).toHaveBeenCalledWith({
    url: "fake-blob-url",
    filename: "activitylogs.json",
  });

  expect(addListener).toHaveBeenCalledTimes(1);
  expect(addListener).toHaveBeenCalledWith(listener);

  expect(removeListener).toHaveBeenCalledTimes(1);
  expect(removeListener).toHaveBeenCalledWith(listener);

  expect(revokeObjectURL).toHaveBeenCalledTimes(1);
  expect(revokeObjectURL).toHaveBeenCalledWith("fake-blob-url");
```

src/lib/Save.js
```diff
@@ -0,0 +1,32 @@
...
+  }
+
+  downloadFile(blob, filename) {
+    const url = URL.createObjectURL(blob);
+
+    function listener(result) {
+      if (result.state.current === 'complete') {
+        setTimeout(() => {
+          URL.revokeObjectURL(url);
+        }, 100);
```
rpl, line 13: @atiqueahmedziad Why are we using a setTimeout here? once we receive the onChanged event for the completed download can't we just immediately revoke that blob url?

src/lib/Save.js
```diff
@@ -0,0 +1,32 @@
...
+        browser.downloads.onChanged.removeListener(listener);
+      }
+    }
+
+    browser.downloads.onChanged.addListener(listener);
+
+    browser.downloads.download({
+      url,
+      filename,
+    });
```
rpl, line 23: [`downloads.download` calls should resolve to a download id](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/downloads/download#Return_value), I think that we may want to use that download id to be sure that the onChanged events are actually related to the download triggered by this downloads.download API call.

Also, based on the docs, there is a chance that downloads.download may reject because the download has been immediately interrupted, and in that case we may want to do some cleanup (e.g. remove the onChanged listener and clear the blob url) and report the error to the caller.

are there other error scenarios that we may need to take into account?

src/lib/Save.js
```diff
@@ -0,0 +1,32 @@
...
+  }
+
+  saveAsJSON() {
```
rpl, line 26: from an API design perspective, could the API method meant to save the logs be an async function? (one that return a promise that is going to be resolved when the download has been completed, or rejected if it did fail)
That should likely make it easier to use by the caller.

src/lib/Load.js
```diff
@@ -0,0 +1,21 @@
+export default class Load {
```
rpl, line 1: could it make sense to keep load and save part of a single module or a single class?

src/lib/Save.js
atiqueahmedziad: I saw somewhere in the internet that setTimeout is recommended for Firefox. But here it totally make sense to revoke that blob url as the download is completed. I will update it. :-)

src/lib/Save.js
atiqueahmedziad: > [`downloads.download` calls should resolve to a download id](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/downloads/download#Return_value), I think that we may want to use that download id to be sure that the onChanged events are actually related to the download triggered by this downloads.download API call.
Yeah, I get a id when the promise is resolved. I will make this chance :-)

> Also, based on the docs, there is a chance that downloads.download may reject because the download has been immediately interrupted, and in that case we may want to do some cleanup (e.g. remove the onChanged listener and clear the blob url) and report the error to the caller.
> 
> are there other error scenarios that we may need to take into account?
Thanks for noting this, going to cover this above corner case in next iteration and I will checkout if there is any other scenarios, we should take care of.

src/lib/Save.js
atiqueahmedziad: Yeah, I agree. :-)

src/lib/Load.js
atiqueahmedziad: Yeah definitely make sense. :-)

atiqueahmedziad:
@rpl  Need a suggestion on this: 
I have updated the [`save log`](https://github.com/mozilla/extension-activity-monitor/pull/2/files#diff-db4a3b16f118709906a0349dbf3e825aR19) code and in current scenario I am saving the `downloadId` like this: 
```
(async () => {
  downloadId = await browser.downloads.download({
      url,
      filename,
  });
})().catch((error) => {
  URL.revokeObjectURL(url);
  browser.downloads.onChanged.removeListener(listener);
  reject(error);
});
```
In the test, I want to return the Id first once the promise resolves and then I want to call the onChanged event with a listener. For that, I wrote like the following:
```
  download.mockImplementation(() => {
    return Promise.resolve(123).then((id) => {
      listener({
        state: { current: 'complete' },
        id,
      });
    });
  });
```
I think the above code is returning the resolved value after it is done with firing the onChanged event with given listener. I am not getting a workaround for returning the resolved value first and then fire the onChanged event. I tried a couple of workaround to fix it, but unable find the solution for it. :-(

Rob--W:
```
let listener;
let downloadPromise;
let downloadDonePromise = new Promise((resolve, reject) => {
  // assign to listener and register listener
});
downloadPromise = await browser.downloads.download(...);
 ...download();
} finally {
```

> In the test, I want to return the Id first once the promise resolves and then I want to call the onChanged event with a listener. For that, I wrote like the following:
> 
> ```
>   download.mockImplementation(() => {
>     return Promise.resolve(123).then((id) => {
>       listener({
>         state: { current: 'complete' },
>         id,
>       });
>     });
>   });
> ```
> 
> I think the above code is returning the resolved value after it is done with firing the onChanged event with given listener. I am not getting a workaround for returning the resolved value first and then fire the onChanged event. I tried a couple of workaround to fix it, but unable find the solution for it. :-(

Your mock does not work as desired, because you are returning the promise created by `.then(...)`.
It seems that your intent was to create and return a promise (`Promise.resolve(123)`), and then invoke the listener when the caller has used it. Your code does not achieve that, for two reasons:

1. You are not returning the initial promise (`Promise.resolve(123)`), but another promise chained to the initial promise (`.then(...)`). The second promise will resolve once the initial promise has resolved AND when the promise handler in `.then(...)` is invoked.
2. Even if you did separate the two, like this:

    ```
    let promise = Promise.resolve(123);
    promise.then(...);
    return promise;
    ```
    ... then it would still not work as expected, because of how promise handlers are queued and handled. In this example, the handler of `.then(...)` would be called before the handler of the caller, because the `.then(...)` is chained to `promise` before the caller was able to do that. If you are interested in a tutorial on scheduling promises, see https://jakearchibald.com/2015/tasks-microtasks-queues-and-schedules/

If you are certain that the invocation of the mock implementation and the registration of the event listener are handled in the same tick of the event loop, then a way to achieve the desired effect is to post a task by delaying the `listener(...)` call with `setTimeout(...)`.

---
## Feedback on `downloadFile`
Some feedback on the organization of the code and the cleanup handlers:

`URL.revokeObjectURL` and `removeListener` are repeated thrice in your `downloadFile` method. To minimize code duplication, the least that you can do is to move those cleanup handlers to one function, or even a `.finally( ... )` handler.

Instead of putting the whole body of `downloadFile` in one big function (i.e. `(resolve, reject) => { ... }`):
 - Declare `downloadFile` as async, and don't return `new Promise(...)`. Instead, use promises in variables and `await` to make it easier to follow the flow of execution.
  - Minimize the body of the `onChanged` listener to only do the minimum that's needed to detect the completion of the download. You can still call resolve/reject when needed.
  - Use `try { /* await download here*/ /* await download completion here */ } finally { /* cleanup here */ }`.

By splitting the `downloadFile` function in smaller pieces, and using `await` + promise variables, it becomes easier to follow the flow of execution, which variables are relevant and what their state is.

atiqueahmedziad:
Thanks a lot @Rob--W for helping and thanks for the reference article(it was very helpful ❤️). I followed your feedbacks and update the files. Would you mind to check if I have done as you asked ?

src/lib/save_load.js
```diff
@@ -0,0 +1,58 @@
+export const load = {
+  readFile(file) {
+    return new Promise((resolve, reject) => {
+      const reader = new FileReader();
+      reader.onload = () => {
+        resolve(reader.result);
+      };
+      reader.onerror = reject;
```
Rob--W, line 8: This would reject with an event object. The error is stored in `reader.error`.

src/lib/save_load.js
```diff
@@ -0,0 +1,58 @@
...
+    });
+
+    try {
+      downloadId = await browser.downloads.download({
+        url,
+        filename,
+      });
+      return await downloadDonePromise;
+    } catch (error) {
+      return error;
```
Rob--W, line 45: Use throw instead, to reject the promise returned by the async function.

src/lib/save_load.js
```diff
@@ -0,0 +1,58 @@
...
+      URL.revokeObjectURL(url);
+      browser.downloads.onChanged.removeListener(listener);
+    }
+  },
+
+  async saveAsJSON(logs) {
+    const blob = new Blob([JSON.stringify(logs)], {
+      type: 'application/json',
+    });
+    return await this.downloadFile(blob, 'activitylogs.json');
```
Rob--W, line 56: Note: The difference between

```
return await promise;
```

and

```
return promise;
```

... is that in the first case, your function will await the promise and be able to catch errors (which is why it is important to use `return await` in the `downloadFile` method),
whereas in the second case the function returns a value (the promise) and any promise rejections will directly be sent to the caller. In the `saveAsJSON` method, it is not necessary to use `await` in `return await ...` because the promise result is not handled but directly passed to the caller.

tests/save_load.test.js
```diff
@@ -0,0 +1,153 @@
...
+  });
+
+  const promiseRemovedListener = new Promise((resolve) => {
+    removeListener.mockImplementation(() => resolve());
+  });
+
+  const isDownloadComplete = await save.saveAsJSON([{ prop1: 'log1' }]);
+
+  await Promise.all([promiseRevokedURL, promiseRemovedListener]);
+
```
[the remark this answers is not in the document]
atiqueahmedziad: since `createObjectURL` has been passed with blob object. I have added a test like the following: 
`expect(createObjectURL).toHaveBeenCalledWith(new Blob());`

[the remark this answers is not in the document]
Rob--W: This doesn't verify the content. I'm surprised that the test would pass, because `expect(createObjectURL).toHaveBeenCalledWith(new Blob());` looks like something that should never pass (`new Blob()` creates a new instance of an opaque object). Does the test really pass if you write it like this?

The test should verify that the file content has been written as expected.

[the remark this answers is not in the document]
atiqueahmedziad: yes, test passes with it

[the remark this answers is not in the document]
atiqueahmedziad: It also passes if I pass the same content,
`expect(createObjectURL).toHaveBeenCalledWith(
    new Blob([JSON.stringify([{ prop1: 'log1' }])])
  );
`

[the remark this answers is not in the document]
Rob--W: Can you try to figure out why the test unexpectedly passes when an obviously non-matching object is passed? Tests would be rather useless if they claim two inequal things to be equal...

E.g. is there a default matcher that causes any Blob to match any other Blob? (the `expect` API is documented here: https://jestjs.io/docs/en/expect.html ).

Ping me on Matrix (or comment here) if you can't figure it out, then I'll take a look.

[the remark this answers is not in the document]
Rob--W: I just looked up Jest's source code, starting from the definition of `toHaveBeenCalledWith`: https://github.com/facebook/jest/blob/4a59daa8715bde6a1b085ff7f4140f3a337045aa/packages/expect/src/spyMatchers.ts#L1130
It is implemented by a function factory at `createToBeCalledWithMatcher`.

... with intern determines argument equality with `isEqualCall` at https://github.com/facebook/jest/blob/4a59daa8715bde6a1b085ff7f4140f3a337045aa/packages/expect/src/spyMatchers.ts#L575

... and `isEqualCall` ultimately uses `equals` to determine equality: https://github.com/facebook/jest/blob/4a59daa8715bde6a1b085ff7f4140f3a337045aa/packages/expect/src/spyMatchers.ts#L62

The logic is implemented in this file: https://github.com/facebook/jest/blob/4a59daa8715bde6a1b085ff7f4140f3a337045aa/packages/expect/src/jasmineUtils.ts#L30

The relevant part of it is that for unknown objects, that it checks whether the internal class is the same (which is the case for comparing Blob and Blob instances, but not Blob and Object instances), and that every **[enumerable](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Enumerability_and_ownership_of_properties)** property has the same value. The `size` and `type` properties of `Blob` instances are not enumerable, so this effectively means that Jest always considers two `Blob` instances equal.

tests/save_load.test.js
```diff
@@ -109,21 +125,22 @@ test('Save saveAsJSON should return error message if error is encountered', asyn
...
+  const log = [{ prop1: 'log1' }];
+
+  await save.saveAsJSON(log).catch((error) => {
+    expect(error.message).toBe('saveerror');
   });
 
   Promise.all([promiseRevokedURL, promiseRemovedListener]);
 
-  expect(download).toHaveBeenCalledTimes(1);
+  expect(createObjectURL).toHaveBeenCalledWith(new Blob([JSON.stringify(log)]));
```
Rob--W, line 136: This check doesn't work as expected for the reason that I provided in https://github.com/mozilla/extension-activity-monitor/pull/2/files/168cec3ae3c1182806ec14f4604ec946dbeffa91..6ac6cf9062aad9f46792ce85bca8f376d3c271ee#r440697786

tests/save_load.test.js
```diff
@@ -153,3 +170,25 @@ test('load data from a file', async () => {
...
+  const file = JSON.stringify([{ prop: 'log' }]);
+  const blob = new Blob([file], {
+    type: 'application/json',
+  });
+
+  readFileFn.mockImplementation(() => {
+    return Promise.reject('loaderror');
+  });
+
+  const loadedLogs = await load.loadLogsStr(blob).catch((error) => {
```
Rob--W, line 187: Use `expect(...).rejects` instead of manual `.catch`: https://jestjs.io/docs/en/expect.html#rejects

There are two issues with the current `.catch` + `expect`:
- It's lots of boilerplate and any resulting logged assertion/expectations are not as obvious.
- If the method doesn't reject, the test passes (when it should have failed instead!)

tests/save_load.test.js
```diff
@@ -0,0 +1,226 @@
...
+    const promiseRemovedListener = new Promise((resolve) => {
+      removeListener.mockImplementation(() => resolve());
+    });
+
+    const log = [{ prop1: 'log1' }];
+
+    const saveAsJson = save.saveAsJSON(log);
+    await expect(saveAsJson).rejects.toThrow('save-error');
+
+    Promise.all([promiseRevokedURL, promiseRemovedListener]);
```
Rob--W, line 201: This has not been addressed yet: https://github.com/mozilla/extension-activity-monitor/pull/2#discussion_r440122183

I think that you misunderstood me. By "don't use `await`", I didn't mean to literally remove `await` (using `Promise.all` without using/awaiting the resulting promise is not useful). I meant that after returning from `readFile`, that the cleanup (revokeObjectURL and removeListener) should have happened already. So you shouldn't use `promiseRevokedURL` (etc., remove it), but verify that the spies have been called as expected (already done below).

tests/save_load.test.js
```diff
@@ -0,0 +1,226 @@
...
+      type: 'application/json',
+    });
+
+    const loadedLogs = await load.loadLogAsText(blob);
+
+    expect(readFileFn).toHaveBeenCalledTimes(1);
+    expect(readFileFn).toBeCalledWith(blob);
+
+    expect(readAsTextFn).toHaveBeenCalledTimes(1);
+    expect(readAsTextFn).toBeCalledWith(blob);
```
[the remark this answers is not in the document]
atiqueahmedziad: Thanks for explaining 😄. Removed them.

tests/save_load.test.js
```diff
@@ -0,0 +1,226 @@
...
+
+    class Reader extends window.FileReader {
+      readAsText() {
+        // error occured during reading
+        super.onerror();
+      }
+    }
+
+    const defaultFileReader = window.FileReader;
+    window.FileReader = Reader;
```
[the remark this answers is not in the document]
atiqueahmedziad: Done :-)

tests/save_load.test.js
```diff
@@ -0,0 +1,226 @@
...
+
+    const loadedLogs = load.loadLogAsText(blob);
+    await expect(loadedLogs).rejects.toThrow();
```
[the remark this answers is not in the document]
atiqueahmedziad: I get a message with string `null` from `FileReader.error`. Now I have added a check for that.

tests/save_load.test.js
```diff
@@ -0,0 +1,226 @@
...
+
+    expect(readFileFn).toHaveBeenCalledTimes(1);
+    expect(readFileFn).toHaveBeenCalledWith(blob);
```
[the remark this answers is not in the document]
atiqueahmedziad: Removed them :-)

tests/save_load.test.js
atiqueahmedziad: Done now :-)

tests/save_load.test.js
```diff
@@ -0,0 +1,163 @@
...
+  test('throws error while reading file', async () => {
+    const readAsTextFn = jest.spyOn(FileReader.prototype, 'readAsText');
+
+    const file = JSON.stringify([{ prop: 'log' }]);
+    const blob = new Blob([file], {
+      type: 'application/json',
+    });
+
+    readAsTextFn.mockImplementationOnce(function () {
+      this.onerror();
```
Rob--W, line 23: In order for the test to be realistic, you should have set a non-empty error matching the type of https://developer.mozilla.org/en-US/docs/Web/API/FileReader/error

atiqueahmedziad: should I pass `DOMException` object as parameter ? Since I am unable to edit the `error` property of FileReader API.

Rob--W: The first parameter is `event`, not `error`. It would be an error to conflate it.

I guess that `error` is a read-only property on the prototype of `FileReader`. Assigning to it won't work for that reason. "read-only" in JavaScript just means that there is a property descriptor with a getter but no setter.

Have you tried to mock the property with the jest API?
If that didn't work, have you tried to use `Object.defineProperty(this, "error", ...)` to set the error on the `FileReader` instance? https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/defineProperty

atiqueahmedziad: Thanks! Updated the pr. I thought I can never access "read-only" properties. Done with jest API.

PR #3, opened 2020-06-15
feat: popup and extension monitoring feature added

atiqueahmedziad:
The following features are added - 
- browerAction popup Added
   - It displays current monitoring status (if any extensions are being monitored or not) (method: `areExtsBeingMonitored()`)
   - Start monitor all extensions
   - Stop monitor all extensions
   - View activitylog page (extension page) 
- Extension monitoring functionalities added
   - Initialize monitor all extensions (method: `initMonitor()`, `setExtensions()`, `startMonitorAll()`)
   - Stop monitor all extensions (method: `stopMonitorAll()`)
   - On the go change any extension's monitoring status (method: `extsCurrentMonitoringStatus()`, `modifyMonitor()`, `startMonitor()`, `stopMonitor()`)
   - Store logs in background (method: `logListener()`)
   - Send all stored logs (in background) to extension page once that is initialized (activityLogs array is send as response when the message - [`sendAllExistingLogs`](https://github.com/mozilla/extension-activity-monitor/compare/master...atiqueahmedziad:popup-monitor?expand=1#diff-961641dd6af66f5e3313307cf4eb8f1dR156-R159) is found)
   - While extension page is opened, send each logs to extension page as soon as it is encountered. (method: `sendLogs()`, `isExtensionPageOpen()`)

**NOTE**: I have added some comments to providing you a better understanding of what I wanted to do. I will take those off after you check

## popup screenshot
Screenshot 2020-06-15 at 3 50 18 PM
Screenshot 2020-06-15 at 3 50 27 PM


src/background.js
```diff
@@ -1 +1,163 @@
-/* background.js */
+const activityLogs = [];
+
+const extensionMonitor = {
+  extensionMapList: new Map([]),
+  isMonitorRunning: false,
+
+  async isExtensionPageOpen() {
+    const tab = await browser.tabs.query({
+      url: ['moz-extension://*/activitylog/activitylog.html'],
```
wagnerand, line 9: Can you use `browser.runtime.getURL()` here, instead of fuzzy-matching using `*`?

src/background.js
```diff
@@ -1 +1,163 @@
-/* background.js */
+const activityLogs = [];
+
+const extensionMonitor = {
```
wagnerand, line 3: This could be its own class.

src/popup/popup.js
```diff
@@ -1 +1,64 @@
...
+  window.close();
+}
+
+async function viewActivityLogs() {
+  browser.tabs.create({
+    url: `${browser.runtime.getURL('../activitylog/activitylog.html')}`,
+  });
+  window.close();
+}
+
```
wagnerand, line 33: Move methods that don't interact with the view to a different class/file. This makes a clearer module structure and allows you to reuse that code.

atiqueahmedziad:
*Note:* **linter will be fixed once pr 2 lands**
## Update from last iteration
- `background.js` codes are being simplified from first iteration.
- Codes are structured in module/ class format.
- Andreas's comment's are checked.
- popup UI updated. (Check the screenshot attached below)
   - When there are no extension being monitored, "Stop Monitoring all Extensions" button is disabled.
   - When extension being monitored, "Start Monitoring all Extensions" button is disabled.

## Current Version 
- browerAction popup modified (all methods imported from `lib/extListen.js`)
    - It displays current monitoring status (if any extensions are being monitored or not) (method: `areExtsBeingMonitored()`)
    - Start monitor all extensions (method: `initMonitorAll()`)
    - Stop monitor all extensions (method: `stopMonitorAll()`)
    - View activitylog page (extension page) (method: `viewActivityLogs()`)

- Extension monitoring functionalities added (all methods imported from `lib/extMonitor.js`)
     - Initialize monitor all extensions (method: `initMonitorAll()`,`setExtensions()`, `startMonitor()`)
     - Stop monitor all extensions (method: `stopMonitorAll()`, `stopMonitor()`)
     - On the go change any extension's monitoring status (method: `modifyMonitor()`, `startMonitor()`, `stopMonitor()`)
     - Store logs in background (method: `logListener()`)
     - Send all stored logs (in background) to extension page once that is initialized (`logs` array from is send as response when the message - sendAllExistingLogs is found)
     - While extension page is opened, send each logs to extension page as soon as it is encountered. (method: `sendLogs()`, `isExtensionPageOpen()`)

## popup screenshot
Screenshot 2020-06-19 at 1 47 32 AM
Screenshot 2020-06-19 at 1 47 50 AM

src/lib/extListen.js
```diff
@@ -0,0 +1,32 @@
+export async function getAllExtensions() {
+  const extensions = await browser.management.getAll();
+  return extensions.filter((extension) => extension.type !== 'theme');
+}
+
```
atiqueahmedziad, line 5: Do any have any suggestion regarding this filename ? (I named it `extListen.js`, as I wasn't getting any good name). This file will contain methods interacts with background script (currently). These methods can be used in both popup and extension page.

atiqueahmedziad: ** Update **: I renamed it to `ext-listen.js`

src/lib/ext-monitor.js
```diff
@@ -0,0 +1,91 @@
...
+    }
+  }
+
+  logListener() {
+    const listener = async (details) => {
+      // Everytime a log is encountered,
+      // it is being pushed to logs array.
+      this.logs.push(details);
+      //it checks if the extension page is open so that it can send log
+      await this.sendLogs(details);
```
atiqueahmedziad, line 29: Here, I am waiting for the promise to be resolved, it is meant to resolve when the extension page is opened. So, it is require to handle the error if promise is reject from [`sendLogs()`](https://github.com/mozilla/extension-activity-monitor/pull/3/files#diff-aa663d2fedefb755dd0cf06fef1f44e2R18) method ?

src/lib/ext-monitor.js
```diff
@@ -0,0 +1,91 @@
...
+
+  setExtensions(extensionsArray) {
+    extensionsArray.forEach((extension) => {
+      this.startMonitor(extension.id);
+    });
+    return Promise.resolve('ext-monitor-started');
+  }
+
+  startMonitor(extensionId) {
+    if (this.extensionMapList.has(extensionId)) {
```
rpl, line 42: I would expect startMonitor to early exit also when extensionId is the id of the extension activity monitor extension ([onExtensionActivity.addListener is also preventing an extension to monitor itself internally, but it also log an error](https://searchfox.org/mozilla-central/rev/1304678d837c17811617754fd446794b774afe94/toolkit/components/extensions/parent/ext-activityLog.js#27-29), which could be misleading if we are investigating a bug in the extension, and so it would be reasonable to prevent the extension from calling onExtensionActivity.addListener with its own extension id in the first place).

src/lib/ext-monitor.js
```diff
@@ -0,0 +1,91 @@
+export default class ExtensionMonitor {
+  logs = [];
+  extensionMapList = new Map([]);
```
rpl, line 3: It would be good to add an inline comment right above this property to explicitly mention what is the type signature of this Map and what it is used for (looking to the rest of the module this seems to be a `Map<string, Function>` where the `string` key is an extension id and the `Function` value is a listener.

src/popup/popup.js
```diff
@@ -1 +1,60 @@
...
+async function init() {
+  const extensions = await getAllExtensions();
+  const startMonitorAllBtn = document.getElementById('startMonitorBtn');
+  const stopMonitorAllBtn = document.getElementById('stopMonitorBtn');
+  const viewActivityLogBtn = document.getElementById('actLogPage');
+  const monitorStatusText = document.getElementById('monitorStatus');
+
+  startMonitorAllBtn.addEventListener('click', async () => {
+    const monitorMsg = await initMonitorAll(extensions);
```
rpl, line 17: In this version of this PR, the popup seems to be retrieving the list of all the extensions installed when it is initialized, but the only purpose of doing it seems to be to send the array of the extensions to the background page.
The background page could just call getAllExtensions internally, and the popup could just send a message to ask the background page to monitor all installed extensions.

src/lib/ext-listen.js
```diff
@@ -0,0 +1,29 @@
+export async function getAllExtensions() {
+  const extensions = await browser.management.getAll();
+  return extensions.filter((extension) => extension.type !== 'theme');
```
rpl, line 3: Personally I would do the opposite and only keep the entries with extension.type == 'extension'.

src/lib/ext-listen.js
```diff
@@ -0,0 +1,29 @@
...
+}
+
+export async function areExtsBeingMonitored() {
+  const { status } = await browser.runtime.sendMessage({
+    getMonitoringStatus: true,
```
rpl, line 8: how about using messages like `{requestType: "getMonitoringStatus"}` instead of `{getMonitoringStatus: true}`?

src/background.html
```diff
@@ -0,0 +1,9 @@
+<!DOCTYPE html>
+<html lang="en">
+  <head>
+    <meta charset="utf-8" />
+    <meta name="viewport" content="width=device-width, initial-scale=1.0" />
```
rpl, line 5: The background page is an invisible page, do we really need this `meta name="viewport"` here?

src/background.html
```diff
@@ -0,0 +1,9 @@
+<!DOCTYPE html>
+<html lang="en">
+  <head>
+    <meta charset="utf-8" />
```
rpl, line 4: Nit, `meta` is a self-closing tag in HTML5 and so we could omit the closing `/` and just write it as `<meta charset="utf-8">`.

src/lib/ext-monitor.js
Rob--W: If anything, I suggest to catch (expected) errors in the `sendLogs` method itself instead of a catch-all in the caller of sendLogs.

In normal circumstances (= tab is open and listening for events) there shouldn't be any error, so I wouldn't catch anything right now.

src/background.js
```diff
@@ -1 +1,38 @@
...
+  // This is send all existing logs to activitylogs page
+  // once activitylogs page is initialized
+  if (sendAllExistingLogs) {
+    sendResponse({
+      existingLogs: extensionMonitor.logs,
+    });
+  }
+}
+
+browser.runtime.onMessage.addListener(messageHandler);
```
rpl, line 38: Why is this part outside of the ExtensionMonitor class?

background.js could just limit itself to create an ExtensionMonitor instance (and maybe call an `init` or a `listen` method to start listening for runtime.onMessage events to handle).

src/background.js
```diff
@@ -1 +1,38 @@
...
+    return extensionMonitor.stopMonitorAll().then((monitorMsg) => monitorMsg);
+  }
+
+  // This is send all existing logs to activitylogs page
+  // once activitylogs page is initialized
+  if (sendAllExistingLogs) {
+    sendResponse({
+      existingLogs: extensionMonitor.logs,
+    });
+  }
```
rpl, line 35: As I did mention in another review comment, these runtime.onMessage events could be better represented by messages in the form:
```
{
  requestType: "...", // where requestType could be "getMonitorStatus", "startMonitor", "stopMonitor" ...
  args: [...], // where args is an array (or an object if you prefer) of arguments for the specific requestType.
}
```

This would:
- make it clear that every one of this messages handles just a single request type
- the message handler could easily delegate the request type to a "requestType"-specific method based on just the `requestType` property.

src/lib/ext-listen.js
```diff
@@ -0,0 +1,29 @@
...
+
+export async function stopMonitorAll() {
+  return await browser.runtime.sendMessage({
+    extStopMonitorAll: true,
+  });
+}
+
+export function viewActivityLogs() {
+  browser.tabs.create({
+    url: `${browser.runtime.getURL('../activitylog/activitylog.html')}`,
```
rpl, line 27: In ext-monitor.js there is a similar `runtime.getURL` (in the `isExtensionPageOpen` method) but with a different url, I'm pretty sure that the urls are different because they are both relative urls and the callers are at a different nesting level in the extension's directory tree. I suggest to use an absolute url (I mean `browser.runtime.getURL('/activitylog/activitylog.html')` and possibly to do it in a single place, e.g. a static method or a function exported by a module used by the other js modules that needs the shared helper) to make it easier to be sure that the url would be resolved to the expected one

src/popup/popup.js
```diff
@@ -1 +1,60 @@
...
+      alert("Extension monitoring couldn't be started");
+    }
+  });
+
+  stopMonitorAllBtn.addEventListener('click', async () => {
+    const monitorMsg = await stopMonitorAll();
+    if (monitorMsg === 'ext-monitor-stopped') {
+      window.close();
+    } else {
+      alert("Extension monitoring couldn't be stopped");
```
rpl, line 30: Personally, I would prefer to avoid alert as the strategy used to show errors to the user.
Also, the popup is being just closed when the handler is successful, which I'm not sure is enough as a feedback to the user that the operation has been successfull (e.g. I would imagine that the user may just open the browserAction popup again to double-check that).

Instead, could we:
- when the handler has successfully started/stopped the monitor extensions: update the popup page (e.g. updating the UI as the user would expect when the button clicked handled the request successfully) and don't autoclose the popup
- when the handler did fail, update the popup page to show an error message to the user and don't autoclose the popup

Besides that, what if initMonitorAll or stopMonitorAll reject?

src/background.js
Rob--W: It doesn't have to be called `args`. The most important part is that using e.g. `requestType` would make it very clear that a message is expected to be handled by one handler only.

There are multiple ways to handle those messages, e.g. `switch`-`case` / `if`-`else if` / putting message handlers on an object (e.g. as seen here: https://github.com/mozilla/FirefoxColor/blob/9d27d0d10249f8b20d36f629b6c9cd7ea14532f0/src/extension/background.js#L46-L108)

src/lib/ext-monitor.js
```diff
@@ -0,0 +1,91 @@
...
+  modifyMonitor(extensionId, isMonitor) {
+    isMonitor ? this.startMonitor(extensionId) : this.stopMonitor(extensionId);
+  }
+
+  getCurrentMonitoredExts() {
+    return this.extensionMapList;
+  }
+
+  areExtsBeingMonitored() {
+    return this.extensionMapList.size > 0 ? true : false;
```
Rob--W, line 82: Don't use ` ? true : false;` when the left part is already a boolean. The comparison itself already results in a boolean.

src/popup/popup.js
```diff
@@ -1 +1,60 @@
...
+  });
+
+  viewActivityLogBtn.addEventListener('click', () => {
+    viewActivityLogs();
+    window.close();
+  });
+
+  const status = await areExtsBeingMonitored();
+
+  if (status) {
```
Rob--W, line 41: All of this status rendering logic should be moved to a separate function, because it is possible to toggle the status while monitoring.

src/lib/ext-monitor.js
```diff
@@ -0,0 +1,91 @@
...
+  }
+
+  initMonitor(extensions) {
+    if (!this.areExtsBeingMonitored()) {
+      return this.setExtensions(extensions);
+    }
+    return Promise.reject(new Error('ext-monitor-init-failed'));
```
Rob--W, line 89: When the function is always expected to return a promise, mark the function as `async` to avoid having to use `Promise.resolve`/`Promise.reject` over `return` / `throw`.

The disadvantage of the current approach is that it is not obvious whether `initMonitor` may throw synchronously (instead of resolving the promise).

src/background.html
atiqueahmedziad: Prettier gives an error on this. I see an issue already filed [here](https://github.com/prettier/prettier/issues/5246) regarding this.
So, I have asked prettier to ignore this for now.

tests/ext-popup.test.js
```diff
@@ -0,0 +1,48 @@
...
+
+test('test init popup when no extension is being monitored', async () => {
+  document.body.innerHTML =
+    '<div>' +
+    '  <p id="monitorStatus"></p>' +
+    '  <button id="startMonitorBtn"></button>' +
+    '  <button id="stopMonitorBtn"></button>' +
+    '  <button id="actLogPage"></button>' +
+    '  <p id="errorText"></p>' +
+    '</div>';
```
rpl, line 12: We could actually load the real document.body from the popup.html file, which would be better for at least a couple of reasons:
- the test would be also make sure that popup.html is what ext-popup.js expects
- we don't have to keep both the test and the popup.html file in sync by manually copying changes between them

To achieve that we could use:
- the nodejs "path" and "fs" modules, to be able to load the file content using something like:
```
  const popupHtml = fs.readFileSync(
    path.resolve(__dirname, "../src/popup/popup.html"),
    "utf-8"
  );
```
- then parse the content of the popup.html file as an HTML document, by creating an instance of [DOMParser](https://developer.mozilla.org/en-US/docs/Web/API/DOMParser) (which is available in jest thanks to JSDOM, available as window.DOMParser) and then using the parseFromString method, it would look like more or less as in: ```domParser.parseFromString(popupHtml, "text/html").body.innerHTML```

(it may be reasonable to do it once outside of the test functions and set it on a `const popupBody`, to eventually reuse it for multiple tests).

tests/ext-popup.test.js
```diff
@@ -0,0 +1,48 @@
...
+  const renderMonitorStartedUIFn = jest.spyOn(popup, 'renderMonitorStartedUI');
+  const addListenerPopupFn = jest.spyOn(popup, 'addListenerPopup');
+
+  areExtsBeingMonitoredFn.mockImplementation(() => {
+    return new Promise((resolve) => resolve(true));
+  });
+
+  await popup.init();
+  expect(addListenerPopupFn).toBeCalledTimes(3);
+  expect(renderMonitorStartedUIFn).toHaveBeenCalledTimes(1);
```
rpl, line 35: I don't think that we are not that much interested into how many times we have called this function, as much as we are interested in checking that the status of the elements in the popup is the expected ones (which is being checked below).

In my opinion we can get read of these two assertions.

tests/ext-popup.test.js
```diff
@@ -0,0 +1,48 @@
+import Popup from '../src/lib/ext-popup';
+import * as ExtListen from '../src/lib/ext-listen';
+
+test('test init popup when no extension is being monitored', async () => {
```
rpl, line 4: This test seems to be testing the popup when the monitor is active, and so if I'm not wrong the description of the test should be changed accordingly.

rpl: @atiqueahmedziad I forgot to ask you this in one of the other review comments: which are the other scenarios that you think you should test on the popup?

Writing down a short list in text form (just as short test descriptions like the one passed to `test("...", async () {...})`) is helpful to plan the tests to write (and would help me to give you feedback on the list of test scenario even before you have turned all the tests in the list into code).

Feel free to reply to this comment with the list of test cases you are thinking of and I'm going to take a look to it asap.

src/lib/ext-monitor.js
atiqueahmedziad: Done. I have excluded monitoring our own extension. :-)

src/lib/ext-monitor.js
atiqueahmedziad: This is done

src/popup/popup.js
atiqueahmedziad: Done :-) Extensions are being fetched in the background now.

src/lib/ext-listen.js
atiqueahmedziad: done :-)

src/lib/ext-listen.js
atiqueahmedziad: This is done. Thanks for the idea.

src/background.html
atiqueahmedziad: no need actually. so removed :-)

src/background.js
atiqueahmedziad: Done.

src/background.js
atiqueahmedziad: Thanks for this idea, I have implemented it.

src/lib/ext-listen.js
atiqueahmedziad: Thanks for noticing, I have added a function named `extPageURL()` in `ext-listen.js` with absolute url. I resued this function two times.

src/popup/popup.js
atiqueahmedziad: Thanks for explaining in detail. I have added a throw statement [here](https://github.com/mozilla/extension-activity-monitor/pull/3/commits/eb10df7504be3bcf734b29a57b6da9df60658294#diff-aa663d2fedefb755dd0cf06fef1f44e2R46) and [here](https://github.com/mozilla/extension-activity-monitor/blob/eb10df7504be3bcf734b29a57b6da9df60658294/src/lib/ext-monitor.js#L95) if initMonitorAll fails.
I have added a throw statement [here](https://github.com/mozilla/extension-activity-monitor/blob/eb10df7504be3bcf734b29a57b6da9df60658294/src/lib/ext-monitor.js#L55) if stopMonitorAll fails. 

Since we are using `try {} catch {}` now. The rejects will be caught and the reject message will be displayed in the popup.

src/lib/ext-monitor.js
atiqueahmedziad: done.

src/popup/popup.js
atiqueahmedziad: done.

src/lib/ext-monitor.js
atiqueahmedziad: Thanks for explaining. I am updated with `throw`.

src/lib/ext-popup.js
```diff
@@ -0,0 +1,93 @@
...
+    initMonitorAll()
+      .then((monitorMsg) => {
+        if (monitorMsg === 'ext-monitor-started') {
+          this.renderPopup();
+        }
+      })
+      .catch((error) => {
+        this.renderErrorMsg(error?.message);
+      });
+  };
```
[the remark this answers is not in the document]
atiqueahmedziad: Yeah, you guessed right. Since I had made the workaround using an arrow function, I just ignored if I could do that in any other way :-p. Also, I didn't know about other workarounds and about the `handleEvent method`. Thanks a lot for the explanation. I have updated this part with your suggestion (added an handleEvent method and used try, catch)

> As a side not given that the reference to the button is called `startMonitorAllBtn` and the other method is called `stopMonitoring`, we could use the same convention can call the methods `startMonitor` and `stopMonitor`.

(I understand you wanted to write `note` instead of `not`). But I am not sure I understand this clearly. I have renamed the methods to `startMonitor` and `stopMonitor`.

src/lib/ext-popup.js
```diff
@@ -0,0 +1,93 @@
...
+    this.monitorStatusText.classList.add('red');
+    this.monitorStatusText.classList.remove('green');
+  }
+
+  async renderPopup() {
+    const isMonitorStarted = await areExtsBeingMonitored();
+    isMonitorStarted
+      ? this.renderMonitorStartedUI()
+      : this.renderMonitorStoppedUI();
+  }
```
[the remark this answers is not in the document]
atiqueahmedziad: done renaming.

src/lib/ext-popup.js
```diff
@@ -0,0 +1,93 @@
...
+  async init() {
+    this.startMonitorAllBtn = document.getElementById('startMonitorBtn');
+    this.stopMonitorAllBtn = document.getElementById('stopMonitorBtn');
+    this.viewActivityLogBtn = document.getElementById('actLogPage');
+    this.monitorStatusText = document.getElementById('monitorStatus');
+    this.errorMsgText = document.getElementById('errorText');
+
+    await this.renderPopup();
+
+    this.addListenerPopup(
```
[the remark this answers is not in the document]
atiqueahmedziad: yeah removed `addListenerPopup` method.

tests/ext-popup.test.js
```diff
@@ -0,0 +1,48 @@
...
+
+  const startBtn = document.getElementById('startMonitorBtn');
+  const stopBtn = document.getElementById('stopMonitorBtn');
+  const monitorStatus = document.getElementById('monitorStatus');
+  const errorText = document.getElementById('errorText');
+
+  const popup = new Popup();
+
+  const renderMonitorStartedUIFn = jest.spyOn(popup, 'renderMonitorStartedUI');
+  const addListenerPopupFn = jest.spyOn(popup, 'addListenerPopup');
```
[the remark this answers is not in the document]
atiqueahmedziad: done.

src/background.html
```diff
@@ -0,0 +1,9 @@
+<!DOCTYPE html>
+<html lang="en">
+  <head>
+    <!-- prettier-ignore -->
+    <meta charset="utf-8">
```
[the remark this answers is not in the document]
atiqueahmedziad: okay :-)

tests/ext-popup.test.js
atiqueahmedziad: Thanks for the explanation. I have updated this part using "path" and "fs" module.

tests/ext-popup.test.js
atiqueahmedziad: Yeah, deleted this two assertions.

tests/ext-popup.test.js
atiqueahmedziad: I want to add the following tests - 

`test("test popup UI when extensions are not being monitored", async () { ... })`
`test("test popup UI when error is thrown while initialize monitoring", async () { ... })`
`test("test popup UI when error is thrown while stop monitoring", async () { ... })`
`test("viewExtPage() method is called upon clicking on 'view activity logs page' button", async () {...}))`

rpl: maybe `"test popup UI when extensions are not being monitored"` could be expanded a bit to test a more complete scenario, e.g. how about:
- `"starting monitoring all extension from the popup"`
  - popup loaded initially in the "not monitoring" state (while we mock the internal helpers to let us simulate this state)
  - check that the UI is in the expected state
  - click the start monitor (after mocking the internal helpers that would let us know that the monitor did start successfully)
  - check that the UI has been updated as expected

how that sounds to you?

rpl: Nit, I would also tweak a bit the description of these two:

- "viewExtPage() method is called upon clicking on 'view activity logs page' button" => "clicking 'view activity logs page' should open the activitylog extension page"
- "test popup UI when an error is thrown while initializing monitoring and stop monitoring" => "errors on start/stopping monitoring should be shown in the popup UI"

(just to express them in terms of the scenario being tested, instead of using a description that is focused on the particular code tested)

atiqueahmedziad: Sounds good. :-) 
Let me expand them in the following - 
- "stop monitoring all the extensions from popup"
    - popup loaded initially in the monitoring state.
    - check that the UI is in the expected state
    - click the stop monitor button
    - check the UI has been updated as expected.
- "error is thrown while initializing monitoring and stop monitoring"
    - popup loaded initially with no error message
    - click on start monitor button/ stop monitor button
    - an error is thrown
    - UI should be updated as expected i.e. the error message is showed.
- "clicking 'view activity logs page' should open the activitylog extension page"
    - popup initially loaded
    - click on "view activity logs page" button
    - it should open a new tab with extension page

src/lib/ext-popup.js
```diff
@@ -0,0 +1,111 @@
...
+  }
+
+  renderErrorMsg(message) {
+    this.errorMsgText.textContent = message;
+    this.errorMsgText.style.display = 'block';
+  }
+
+  renderMonitorStartedUI() {
+    this.startMonitorAllBtn.setAttribute('disabled', 'disabled');
+    this.startMonitorAllBtn.classList.add('disabled');
```
rpl, line 24: Nit, we can avoid to add the disabled class by using the pseudo classes in the CSS (See the MDN doc page for the [:disabled](https://developer.mozilla.org/en-US/docs/Web/CSS/:disabled) pseudo class)

src/lib/ext-popup.js
```diff
@@ -0,0 +1,111 @@
...
+    this.stopMonitorAllBtn.removeAttribute('disabled');
+    this.stopMonitorAllBtn.classList.remove('disabled');
+
+    this.monitorStatusText.textContent = 'Extensions are being monitored';
+    this.monitorStatusText.classList.add('green');
+    this.monitorStatusText.classList.remove('red');
```
rpl, line 30: Here, it would be better to use class names that reflects the meaning of the state the element should represent instead of the particular style being currently used, in other words: `"green"` could be `"success"` and `"red"` could be `"failure"` (or something like that).

This way if we change ideas about which style to apply, we don't need to change the class names being assigned to the elements.

src/lib/ext-popup.js
```diff
@@ -0,0 +1,111 @@
...
+
+  viewExtPagePopup() {
+    viewExtPage();
+    window.close();
+  }
+
+  async handleEvent(event) {
+    if (event.type === 'click') {
+      switch (event.target.id) {
+        case 'startMonitorBtn':
```
rpl, line 61: Nit, any reasons to use the ids of the elements instead of the references to the DOM elements that the class instance does set as properties in the constructor?

src/lib/ext-popup.js
```diff
@@ -0,0 +1,111 @@
...
+      }
+    } else {
+      throw new Error('wrong event found');
+    }
+  }
+
+  async startMonitor() {
+    try {
+      const monitorMsg = await startMonitorAll();
+      if (monitorMsg === 'ext-monitor-started') {
```
rpl, line 81: Nit , why not just resolving to a boolean so that we don't need to compare it with a particular string value?
(same comment on stopMonitor).

src/lib/ext-popup.js
```diff
@@ -0,0 +1,111 @@
...
+        await this.render();
+      } else {
+        throw new Error("Monitoring couldn't be started");
+      }
+    } catch (error) {
+      this.renderErrorMsg(error?.message);
```
rpl, line 87: Nit, when is the `error` object going to be undefined here?
(same comment on stopMonitor).

src/popup/popup.html
```diff
@@ -1 +1,19 @@
...
+  <body>
+    <div class="container">
+      <p id="monitorStatus"></p>
+      <button id="startMonitorBtn">Start monitoring all extensions</button>
+      <button id="stopMonitorBtn">Stop monitoring all extensions</button>
+      <button id="actLogPage">View activity logs page</button>
+      <p id="errorText"></p>
+    </div>
+
+    <script src="popup.js" type="module"></script>
```
rpl, line 17: In the test we are currently loading this HTML file to inject it into the JSDOM document and test the behaviors we expect, in that case we do not need (and we do not want) JSDOM to load this tag script because in the test we do create the Popup instance as part of the test.

To avoid loading the script tags when we do inject the content of the popup body as part of the test we could move the script tag into the head and add a `defer` attribute to it (so that it will basically behave as it does when the script tag is at the end of the body tag content).

Looking to the docs for the [`defer` attribute on MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/script#attr-defer) it looks that we don't even need to explicitly set the `defer` attribute of this `script` tag because it is `type="module` and it should be deferred automatically (which means that it would be executed after the document has been parsed, but before DOMContentLoad is fired, which is what we need given that popup.js is listening for that event and Popup class constructor does expect to be able to retrieve the DOM elements that are part of the document body).

atiqueahmedziad: Thanks for explaining this.

src/lib/ext-popup.js
atiqueahmedziad: it will not. :-) so removed the optional.

src/lib/ext-popup.js
atiqueahmedziad: Yeah made it boolean.

src/lib/ext-popup.js
atiqueahmedziad: using it now :-) Thanks for noticing.

src/lib/ext-popup.js
atiqueahmedziad: Yeah, thanks.

src/lib/ext-popup.js
atiqueahmedziad: Done. :-)

src/lib/ext-monitor.js
```diff
@@ -0,0 +1,127 @@
...
+      return this.setExtensions(extensions);
+    }
+    throw new Error('EAM is already running');
+  }
+
+  messageHandlers = {
+    getMonitorStatus: () => {
+      return Promise.resolve({
+        status: this.areExtsBeingMonitored(),
+      });
```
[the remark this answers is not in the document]
atiqueahmedziad: I am not been able to get the object if I do 
```
return {
     active: this.hasActivityListeners(),
 };
```
Did you mean to do this ?

tests/ext-popup.test.js
```diff
@@ -0,0 +1,246 @@
...
+  expect(openActivityLogPageFn).toHaveBeenCalled();
+
+  // nothing changes from previous check
+  expect(stopBtn.getAttribute('disabled')).toEqual('disabled');
+  expect(startBtn.getAttribute('disabled')).toBeNull();
+
+  expect(monitorStatus.textContent).toBe('No extensions are being monitored');
+  expect(monitorStatus.classList.contains('failure')).toBeTruthy();
+  expect(monitorStatus.classList.contains('success')).toBeFalsy();
+  expect(errorText.textContent).toBe('');
```
rpl, line 245: This is a very long test to only test a small number of lines :-)

We could rip it out or at least try to make it smaller, e.g. all this part is checking something that is already being tested in another test case, and in this test we are only interested in checking that the "view logs" button is going to open the extension page, and so we could reduce it a lot by just doing and checking the bare minimum, e.g.
```
  const viewExtPageBtn = popup.viewActivityLogBtn;
  getMonitorStatusFn.mockResolvedValue(false);
  await popup.init();
  viewExtPageBtn.click();
  expect(openActivityLogPageFn).toHaveBeenCalled();
```

Besides that, (in general not just related to this particular test case) don't worry too much about getting the full coverage for now, we are not going to block the PR until the coverage is 100%, as long as the code is nice and clean and does what it should in the most simple way possible we are going to merge the PR and move forward.

Writing these unit tests is currently meant to help you to:

- learn more about testing (which is always useful)
- re-look to your code from other perspectives, find issues that you may have missed (and eventually test explicitly the part that matter the most)

tests/ext-popup.test.js
```diff
@@ -0,0 +1,246 @@
...
+
+  const startBtn = popup.startMonitorAllBtn;
+  const stopBtn = popup.stopMonitorAllBtn;
+  const monitorStatus = popup.monitorStatusText;
+  const errorText = popup.errorMsgText;
+
+  // initially there will be no monitored extensions
+  getMonitorStatusFn.mockResolvedValueOnce(false);
+  getMonitorStatusFn.mockResolvedValueOnce(true);
+  startMonitorFn.mockResolvedValue('ext-monitor-started');
```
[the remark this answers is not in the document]
atiqueahmedziad: Yeah, sorry I missed it. Fixed now.

src/lib/ext-popup.js
```diff
@@ -0,0 +1,107 @@
...
+          break;
+        default:
+          throw new Error('wrong event target found ' + event.target);
+      }
+    } else {
+      throw new Error('wrong event type found ' + event.type);
+    }
+  }
+
+  async startMonitor() {
```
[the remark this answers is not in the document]
atiqueahmedziad: Done :-)

src/lib/ext-listen.js
```diff
@@ -0,0 +1,36 @@
...
+    requestType: 'getMonitorStatus',
+  });
+  return active;
+}
+
+export async function startMonitor() {
+  return await browser.runtime.sendMessage({
+    requestType: 'startMonitor',
+  });
+}
```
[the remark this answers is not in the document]
atiqueahmedziad: done

src/lib/ext-listen.js
```diff
@@ -0,0 +1,36 @@
...
+  return browser.runtime.getURL('/activitylog/activitylog.html');
+}
+
+export async function getAllExtensions() {
+  const extensions = await browser.management.getAll();
+  const self = await browser.management.getSelf();
+  return extensions.filter((extension) => {
+    return extension.type === 'extension' && extension.id !== self.id;
+  });
+}
```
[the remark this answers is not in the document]
atiqueahmedziad: done.

tests/ext-listen.test.js
```diff
@@ -0,0 +1,100 @@
...
+  });
+
+  const filteredExts = demoExtensions.filter((ext) => {
+    return ext.type === 'extension' && ext.id !== selfId;
+  });
+
+  const extensionsPromise = ExtListen.getAllExtensions();
+  await expect(extensionsPromise).resolves.toMatchObject(filteredExts);
+  expect(getAll).toHaveBeenCalled();
+});
```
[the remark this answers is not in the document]
atiqueahmedziad: Thanks for explaining. I have updated the other test explanation too (where I felt necessary).

tests/ext-popup.test.js
atiqueahmedziad: Thanks for explaining :-)

src/lib/ext-monitor.js
```diff
@@ -0,0 +1,110 @@
...
+    },
+    stopMonitor: () => {
+      return this.stopMonitor();
+    },
+    sendAllLogs: () => {
+      return {
+        existingLogs: this.logs,
+      };
+    },
+  };
```
rpl, line 95: At the moment all this seems to be simple enough to be rolled into a `switch` statement in messageListener, but I'm fine if we want to keep it as it is for now and rethink about it when we are adding more pieces of the architecture.

Nit, if we are keeping it as it, all these could be one lines, by writing them like:
```js
  messageHandlers = {
    getMonitorStatus: () => ({ active: this.hasActivityListeners() }),
    startMonitor: () => this.startMonitor(),
    stopMonitor: () => this.stopMonitor(),
    sendAllLogs: () => ({ existingLogs: this.logs }),
  };
```

src/lib/ext-monitor.js
```diff
@@ -0,0 +1,110 @@
...
+
+  messageListener = async (message) => {
+    const { requestType } = message;
+
+    if (requestType in this.messageHandlers) {
+      return this.messageHandlers[requestType]();
+    } else {
+      throw new Error('requestType ' + requestType + ' is not found');
+    }
+  };
```
rpl, line 105: I've been thinking about this a bit more after I suggested you that this could have been an async function yesterday:

at the moment this may be fine because this is the only `browser.runtime.onMessage` listener that the extension is registering, but at some point this may become confusing because the WebExtensions internals will assume that this messageListener is always handling any message it receives (because it always return a promise).

We may consider this a follow up to take care of in one of the next pull request, but let's keep it in mind.

To avoid this we may do something like this:
- do not make this function as async
- include another property that make it explicitly which module is supposed to handle the message (e.g. a `requestTo` property)
- return earlier if the message wasn't meant to be handled by this message listener

which would look like more or less like this:
```
messageListener = async (message) => {
  const { requestType, requestTo } = message;
  if (requestTo !== "ext-monitor) {
    return;
  }

  if (requestType in this.messageHandlers) {
    try {
      return Promise.resolve(this.messageHandlers[requestType]());
    } catch (err) {
      return Promise.reject(err);
    }
  } else {
    return Promise.reject(new Error('...'));
  }
}
```

and then include `requestTo: 'ext-monitor'` in the methods defined in ext-listen.js that are sending the control message that this module should handle.

src/lib/ext-popup.js
```diff
@@ -0,0 +1,105 @@
...
+    this.errorMsgText = document.getElementById('errorText');
+  }
+
+  renderErrorMsg(message) {
+    this.errorMsgText.textContent = message;
+    this.errorMsgText.style.display = 'block';
+  }
+
+  renderMonitorStartedUI() {
+    this.startMonitorAllBtn.setAttribute('disabled', 'disabled');
```
rpl, line 23: Nit, setting the value to 'disabled' is fine, but I usually prefer to set it to `true`.

src/lib/ext-popup.js
```diff
@@ -0,0 +1,105 @@
...
+  }
+
+  async handleStartMonitor() {
+    try {
+      const monitorMsg = await startMonitor();
+      if (monitorMsg) {
+        this.render();
+      } else {
+        throw new Error("Monitoring couldn't be started");
+      }
```
rpl, line 79: Nit, It doesn't seem that startMonitor can actually ever return false (it rejects if it can't start the monitor because it is already active), and so this may be likely simplified to just do:
```js
try {
  await startMonitor();
  this.render();
} catch (error) {
  this.renderErrorMsg(error.message);
}
```

same in handleStopMonitor.

tests/ext-popup.test.js
```diff
@@ -0,0 +1,204 @@
...
+
+  getMonitorStatusFn.mockResolvedValueOnce(false);
+  getMonitorStatusFn.mockResolvedValueOnce(true);
+  startMonitorFn.mockResolvedValue(true);
+
+  await popup.init();
+
+  // popup at initial - no extensions being watched
+  expect(startBtn.getAttribute('disabled')).toBeNull();
+  expect(stopBtn.getAttribute('disabled')).toEqual('disabled');
```
rpl, line 46: Nit, as an alternative this assertion could use [`hasAttribute`](https://developer.mozilla.org/en-US/docs/Web/API/Element/hasAttribute) or assert the value set on the [InputElement's `disabled` property](https://developer.mozilla.org/en-US/docs/Web/API/HTMLInputElement) (no really needed to changed, I just wanted to mention it)

atiqueahmedziad: Thanks for telling this. I think `hasAttribute` is more appropriate in this case, so I made these changes with `hasAttribute`

PR #4, opened 2020-06-15
feat: activitylog extension page added

atiqueahmedziad:
In current activitylog page (extension page): 
- Display existing logs captured in background at initial loading. (method: `getExistingLogs()`)
- As long as the extension page is not destroyed, every logs captured in background will be passed to the extension page for display. (upon receiving message `eam_updateLogs`, every new log to activityLogs array)
- Logs can be saved as json file.
- ~~Logs can be loaded as json file.~~ (deferred to a separate pull request)

**NOTE**: There is a lot to clean up in the code. For now, it just serves the above purposes. I will clean it up very soon.

### DIFF
--- a/src/activitylog/activitylog.css
+++ b/src/activitylog/activitylog.css
@@ -1,1 +1,120 @@
-/* activitylog.css */
+*,
+*:before,
+*:after {
+  padding: 0;
+  margin: 0;
+  box-sizing: border-box;
+}
+
+.container {
+  width: 100%;
+}
+
+#showActivityLog,
+#showLoadedLog,
+#saveLogBtn {
+  margin-top: 20px;
+}
+
+.activity-log-wrapper {
+  position: relative;
+}
+
+.show-log-detail-wrapper {
+  position: absolute;
+  top: 0;
+  height: 500px;
+  right: 0;
+  width: 40%;
+  display: none;
+  word-wrap: anywhere;
+  background-color: #ffffff;
+  border: 2px solid #000000;
+}
+
+#showLogDetails {
+  margin: 50px 15px 10px;
+}
+
+.log-table table {
+  width: 100%;
+}
+
+.log-table table tbody tr {
+  text-align: center;
+}
+
+table {
+  width: 100%;
+  border: 1px solid #eeeeee;
+}
+
+th {
+  width: 100%;
+  background: #000;
+  padding: 5px;
+  text-transform: uppercase;
+  color: #ffffff;
+}
+
+tr {
+  display: flex;
+  width: 100%;
+}
+
+tr:nth-of-type(odd) {
+  background: #eeeeee;
+}
+
+td,
+th {
+  flex: 1 1 20%;
+  text-align: center;
+  word-wrap: anywhere;
+  align-self: center;
+}
+
+table td {
+  padding: 3px;
+}
+
+.failure {
+  background-color: #961616;
+  padding: 10px;
+  color: #ffffff;
+}
+
+.log-table {
+  overflow-y: scroll;
+  max-height: 500px;
+  height: auto;
+}
+
+table tr:first-child {
+  position: sticky;
+  top: 0;
+}
+
+table {
+  border-collapse: collapse;
+  width: 100%;
+  cursor: pointer;
+}
+
+.width-60 {
+  width: 60%;
+}
+
+.close {
+  background: url(/icons/close.svg) no-repeat center center / cover;
+  position: absolute;
+  width: 40px;
+  height: 40px;
+  top: 5px;
+  right: 5px;
+  cursor: pointer;
+}
+
+.activity-log-container {
+  position: relative;
+}
--- a/src/activitylog/activitylog.css
+++ b/src/activitylog/activitylog.css
@@ -25,7 +25,6 @@
   height: 500px;
   right: 0;
   width: 40%;
-  display: none;
   word-wrap: anywhere;
   background-color: #ffffff;
   border: 2px solid #000000;
--- a/src/activitylog/activitylog.css
+++ b/src/activitylog/activitylog.css
@@ -11,9 +11,8 @@
 }
 
 #showActivityLog,
-#showLoadedLog,
 #saveLogBtn {
-  margin-top: 20px;
+  margin: 20px 0;
 }
 
 .activity-log-wrapper {
@@ -38,10 +37,6 @@
 
 .log-table table {
   width: 100%;
-}
-
-.log-table table tbody tr {
-  text-align: center;
 }
 
 table {
@@ -90,7 +85,7 @@ table td {
   height: auto;
 }
 
-table tr:first-child {
+table thead {
   position: sticky;
   top: 0;
 }
--- a/src/activitylog/activitylog.html
+++ b/src/activitylog/activitylog.html
@@ -1,1 +1,46 @@
-<!-- activitylog.html -->
+<!DOCTYPE html>
+<html lang="en">
+  <head>
+    <meta charset="UTF-8" />
+    <meta name="viewport" content="width=device-width, initial-scale=1.0" />
+    <title>Extension Activity Monitor</title>
+    <link rel="stylesheet" href="activitylog.css" />
+    <script src="activitylog.js" type="module"></script>
+  </head>
+  <body>
+    <div class="container">
+      <h1>Extension Activity Monitor</h1>
+      <div class="load-file-wrapper">
+        <input id="logInputFile" type="file" accept="application/json" />
+        <button id="loadLogBtn" disabled>Load logs</button>
+      </div>
+      <button id="saveLogBtn">Save logs</button>
+      <div class="notice"></div>
+      <div class="activity-log-wrapper">
+        <h2>Activity Logs</h2>
+        <div class="activity-log-container">
+          <div class="log-table">
+            <table>
+              <tr>
+                <th>Extension Id</th>
+                <th>Timestamp</th>
+                <th>API Type</th>
+                <th>API Name</th>
+                <th>View Type</th>
+              </tr>
+              <tbody></tbody>
+            </table>
+          </div>
+          <div class="show-log-detail-wrapper">
+            <div id="showLogDetails"></div>
+            <div class="close"></div>
+          </div>
+        </div>
+      </div>
+      <div class="load-log-wrapper">
+        <h2>Loaded Logs</h2>
+        <div id="showLoadedLog"></div>
+      </div>
+    </div>
+  </body>
+</html>
--- a/src/activitylog/activitylog.html
+++ b/src/activitylog/activitylog.html
@@ -29,7 +29,7 @@
               <tbody></tbody>
             </table>
           </div>
-          <div class="show-log-detail-wrapper">
+          <div class="show-log-detail-wrapper" hidden>
             <div id="showLogDetails"></div>
             <div class="close"></div>
           </div>
--- a/src/activitylog/activitylog.html
+++ b/src/activitylog/activitylog.html
@@ -10,10 +10,6 @@
   <body>
     <div class="container">
       <h1>Extension Activity Monitor</h1>
-      <div class="load-file-wrapper">
-        <input id="logInputFile" type="file" accept="application/json" />
-        <button id="loadLogBtn" disabled>Load logs</button>
-      </div>
       <button id="saveLogBtn">Save logs</button>
       <div class="notice"></div>
       <div class="activity-log-wrapper">
@@ -21,13 +17,15 @@
         <div class="activity-log-container">
           <div class="log-table">
             <table>
-              <tr>
-                <th>Extension Id</th>
-                <th>Timestamp</th>
-                <th>API Type</th>
-                <th>API Name</th>
-                <th>View Type</th>
-              </tr>
+              <thead>
+                <tr>
+                  <th>Extension Id</th>
+                  <th>Timestamp</th>
+                  <th>API Type</th>
+                  <th>API Name</th>
+                  <th>View Type</th>
+                </tr>
+              </thead>
               <tbody></tbody>
             </table>
           </div>
@@ -36,10 +34,6 @@
             <div class="close"></div>
           </div>
         </div>
-      </div>
-      <div class="load-log-wrapper">
-        <h2>Loaded Logs</h2>
-        <div id="showLoadedLog"></div>
       </div>
     </div>
   </body>
--- a/src/activitylog/activitylog.js
+++ b/src/activitylog/activitylog.js
@@ -1,5 +1,5 @@
 import ActivityLog from '../lib/ext-activitylog.js';
 
-window.addEventListener('DOMContentLoaded', async () => {
+document.addEventListener('DOMContentLoaded', async () => {
   await new ActivityLog().init();
 });
--- a/src/activitylog/activitylog.js
+++ b/src/activitylog/activitylog.js
@@ -1,1 +1,5 @@
-/* activitylog.js */
+import ActivityLog from '../lib/ext-activitylog.js';
+
+window.addEventListener('DOMContentLoaded', async () => {
+  await new ActivityLog().init();
+});
--- a/src/lib/ext-activitylog.js
+++ b/src/lib/ext-activitylog.js
@@ -90,9 +90,7 @@ export default class ActivityLog {
           this.closeDetailSidebar();
           break;
         default:
-          throw new Error(
-            `unexpected click event on ${JSON.stringify(event.target.tagName)}`
-          );
+          throw new Error(`unexpected click event on ${event.target.tagName}`);
       }
     } else {
       throw new Error(`wrong event type found ${event.type}`);
@@ -114,7 +112,7 @@ export default class ActivityLog {
   async saveLogs() {
     try {
       await save.saveAsJSON(this.logs);
-      this.setError(false);
+      this.setError(null);
     } catch (error) {
       this.setError(error.message);
     }
--- a/src/lib/ext-activitylog.js
+++ b/src/lib/ext-activitylog.js
@@ -1,0 +1,143 @@
+import { save, load } from '../lib/save-load.js';
+
+export default class ActivityLog {
+  constructor() {
+    this.logs = [];
+    this.loadedLogs = [];
+    this.rowIndex = 0;
+
+    this.loadLogBtn = document.getElementById('loadLogBtn');
+    this.logInputFile = document.getElementById('logInputFile');
+    this.saveLogBtn = document.getElementById('saveLogBtn');
+    this.showLogDetailWrapper = document.querySelector(
+      '.show-log-detail-wrapper'
+    );
+    this.showLogDetails = document.getElementById('showLogDetails');
+    this.showLoadedLog = document.getElementById('showLoadedLog');
+    this.logTableWrapper = document.querySelector('.log-table');
+    this.table = document.querySelector('table');
+    this.notice = document.querySelector('.notice');
+    this.closeBtn = document.querySelector('.close');
+  }
+
+  async getExistingLogs() {
+    const { existingLogs } = await browser.runtime.sendMessage({
+      requestType: 'sendAllLogs',
+      requestTo: 'ext-monitor',
+    });
+    return existingLogs;
+  }
+
+  async loadFile(file) {
+    try {
+      const loadedLogsAsText = await load.loadLogAsText(file);
+      const loadedJSONLogs = JSON.parse(loadedLogsAsText);
+      this.loadedLogs.push(...loadedJSONLogs);
+      // TODO: Display loaded logs in table view
+
+      if (this.notice.textContent.trim().length) {
+        this.notice.textContent = '';
+        this.notice.classList.remove('failure');
+      }
+    } catch (error) {
+      this.notice.textContent = error.message;
+      this.notice.classList.add('failure');
+    }
+  }
+
+  async saveLogs(logs) {
+    try {
+      await save.saveAsJSON(logs);
+
+      if (this.notice.textContent.trim().length) {
+        this.notice.textContent = '';
+        this.notice.classList.remove('failure');
+      }
+    } catch (error) {
+      this.notice.textContent = error.message;
+      this.notice.classList.add('failure');
+    }
+  }
+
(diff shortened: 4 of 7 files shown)


src/lib/ext-activitylog.js
```diff
@@ -0,0 +1,109 @@
...
+  };
+
+  renderTable(logs) {
+    for (const log of logs) {
+      const newRow = document.createElement('tr');
+      newRow.insertAdjacentHTML('beforeend', `<td>${log.id}</td>`);
+      newRow.insertAdjacentHTML('beforeend', `<td>${log.timeStamp}</td>`);
+      newRow.insertAdjacentHTML('beforeend', `<td>${log.type}</td>`);
+      newRow.insertAdjacentHTML('beforeend', `<td>${log.name}</td>`);
+      newRow.insertAdjacentHTML('beforeend', `<td>${log.viewType}</td>`);
```
rpl, line 57: @atiqueahmedziad As we discussed yesterday during our weekly meeting, DOM API methods that accept non sanitized HTML content should be avoided, because they would open the monitor extension to XSS attacks originated by one of the monitored extensions.

Let's remove these ones (e.g. by creating the `td` elements and then filling them with the expected text content using the textContent property).

As I did mention in the meeting we may also add [eslint-plugin-no-unsanitized](https://github.com/mozilla/eslint-plugin-no-unsanitized) to the dev dependencies and then enable it in the eslint configuration used in this repository, as an additional way to prevent this kind of dangerous code pattern to slip into a pull request without being noticed.

src/lib/ext-activitylog.js
```diff
@@ -0,0 +1,160 @@
...
+
+  renderTable(logs) {
+    for (const log of logs) {
+      const newRow = this.table.insertRow(1);
+      newRow.insertCell(0).textContent = log.id;
+      newRow.insertCell(1).textContent = log.timeStamp;
+      newRow.insertCell(2).textContent = log.type;
+      newRow.insertCell(3).textContent = log.name;
+      newRow.insertCell(4).textContent = log.viewType || 'undefined';
+      newRow.setAttribute('index', this.logs.length);
```
atiqueahmedziad, line 103: To make this less complicated, I am not using `rowIndex` property of table's row. Instead, I am adding an attribute. Since we are using `Array.push` to store logs and table rows are being generated in the opposite way (the latest rows stays at top). I can make the adjustment using `Array.unsift` and use `rowIndex`, if you say. :-)

src/lib/ext-activitylog.js
atiqueahmedziad: Thanks for mentioning again. I have opened pr #8 to add "eslint-plugin-no-unsanitized". :-)

src/lib/ext-activitylog.js
```diff
@@ -0,0 +1,160 @@
...
+    });
+    return existingLogs;
+  }
+
+  async loadFile(file) {
+    try {
+      const loadedLogsAsText = await load.loadLogAsText(file);
+      const loadedJSONLogs = JSON.parse(loadedLogsAsText);
+      this.loadedLogs.push(...loadedJSONLogs);
+      this.showLoadedLog.textContent = loadedLogsAsText;
```
rpl, line 34: The part that loads the logs from file seems to be a bit "out of place" at the moment, it doesn't really do much in this pull request and what it does isn't really that useful yet (e.g. the logs loaded are not rendered as rows in the table)

We may instead defer "load from file" to a separate pull request and think a bit more "what load from file should be really doing") and have a better and more clear plan about it, e.g.:
how is the activity log page going to serve this feature to the user?
how is the user going to switch between looking to logs loaded from file and logs being collected in that session from the monitored extensions? how is the activity log page going to may make it clear to the user what logs is looking to (loaded from file or collected in realtime)? and other similar design questions.

atiqueahmedziad: Yeah, I have comment out a TODO here, which I will take care of in the next pr. :-) In the meantime, I will make a clear design answering the above questions.

src/lib/ext-activitylog.js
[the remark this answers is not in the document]
atiqueahmedziad: Since I am pushing the logs at once and then calling this method, for which this is occurring. Fixed it now. :-)

src/lib/ext-activitylog.js
```diff
@@ -0,0 +1,160 @@
...
+      let logIndex;
+      switch (event.target.tagName) {
+        case 'TD':
+          logIndex = event.target.parentElement.attributes['index'].value - 1;
+          this.renderSideBar(logIndex);
+          break;
+        case 'TR':
+          logIndex = event.target.attributes['index'].value - 1;
+          this.renderSideBar(logIndex);
+          break;
```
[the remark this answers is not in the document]
atiqueahmedziad: Thanks. It helped me in debugging.
>  - another arbitrary element that is outside of a row (and in this case we expect closest to return null)
I am not been able to reproduce this, but I have added a condition for `null` and `undefined` value in the code.

src/lib/ext-activitylog.js
```diff
@@ -0,0 +1,160 @@
...
+    this.showLogDetails.textContent = logStringfy;
+    this.showLogDetailWrapper.style.display = 'block';
+  }
+
+  messageHandlers = {
+    appendLogs: ({ log }) => {
+      this.logs.push(log);
+      this.renderTable([log]);
+    },
+  };
```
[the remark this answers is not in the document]
atiqueahmedziad: Yeah, named it `renderNewRow`, which seems meaning for this I guess.

src/lib/ext-activitylog.js
```diff
@@ -0,0 +1,160 @@
...
+    )[0];
+    this.showLogDetails = document.getElementById('showLogDetails');
+    this.showLoadedLog = document.getElementById('showLoadedLog');
+    this.table = document.getElementsByTagName('table')[0];
+    this.tableFirstRow = document.querySelector('tbody tr:first-child');
+    this.notice = document.getElementsByClassName('notice')[0];
+  }
+
+  logs = [];
+  loadedLogs = [];
```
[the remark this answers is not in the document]

src/lib/ext-activitylog.js
```diff
@@ -0,0 +1,160 @@
+import { save, load } from '../lib/save-load.js';
+
+export default class ActivityLog {
+  constructor() {
+    this.loadLogBtn = document.getElementById('loadLogBtn');
+    this.logInputFile = document.getElementById('logInputFile');
+    this.saveLogBtn = document.getElementById('saveLogBtn');
+    this.showLogDetailWrapper = document.getElementsByClassName(
+      'show-log-detail-wrapper'
+    )[0];
```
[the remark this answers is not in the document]

src/activitylog/activitylog.html
```diff
@@ -1 +1,44 @@
...
+  <body>
+    <div class="container">
+      <h1>Extension Activity Monitor</h1>
+      <div class="load-file-wrapper">
+        <input id="logInputFile" type="file" accept="application/json" />
+        <button id="loadLogBtn" disabled>Load logs</button>
+      </div>
+      <button id="saveLogBtn">Save logs</button>
+      <div class="notice"></div>
+      <div class="activty-log-wrapper">
```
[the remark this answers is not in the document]
atiqueahmedziad: sorry for this, I should have been careful with typos. Thanks for noticing. :-)

src/lib/ext-activitylog.js
```diff
@@ -0,0 +1,143 @@
...
+    });
+    return existingLogs;
+  }
+
+  async loadFile(file) {
+    try {
+      const loadedLogsAsText = await load.loadLogAsText(file);
+      const loadedJSONLogs = JSON.parse(loadedLogsAsText);
+      this.loadedLogs.push(...loadedJSONLogs);
+      // TODO: Display loaded logs in table view
```
rpl, line 36: Personally I think that we may defer it to a separate follow up where we can focus on it, and keep this one simpler and more focused on the logs table view.

Anyway, given that I did take a look to it, I guess that we shouldn't accumulate logs loaded from unrelated log files (at the moment I'm not even sure if this class should have both the `this.logs` and `this.loadedLogs` array, because it isn't clear who will be responsible to handle the loaded logs, where they will be rendered etc.).

src/lib/ext-activitylog.js
```diff
@@ -0,0 +1,143 @@
...
+
+      this.renderSideBar(selectedRowIndex);
+    } else {
+      throw new Error('wrong event type found ' + event.type);
+    }
+  }
+
+  renderNewRow(logs) {
+    for (const log of logs) {
+      const newRow = this.table.insertRow(1);
```
rpl, line 86: It is honestly the first time that I see insertRow and insertCell being used :-P
I think that it's fine to use it for now,  if we change idea (or we have other needs) we may change this later.

How is this table ordered? I was expecting this to be this.table.insertRow(-1) which should append it to the end of the table.

src/lib/ext-activitylog.js
```diff
@@ -0,0 +1,143 @@
...
+      newRow.insertCell(0).textContent = log.id;
+      newRow.insertCell(1).textContent = log.timeStamp;
+      newRow.insertCell(2).textContent = log.type;
+      newRow.insertCell(3).textContent = log.name;
+      newRow.insertCell(4).textContent = log.viewType || 'undefined';
+      newRow.setAttribute('index', this.rowIndex++);
+
+      newRow.addEventListener('click', this);
```
rpl, line 94: Instead of registering the class as an event handler on each row, I think we may just do it once in the `init` method on the entire table by doing `this.table.addEventListener('click', this);`.

src/lib/ext-activitylog.js
```diff
@@ -0,0 +1,143 @@
...
+        this.loadLogBtn.setAttribute('disabled', true);
+      }
+    });
+
+    this.closeBtn.addEventListener('click', () => {
+      if (this.showLogDetailWrapper.style.display === 'block') {
+        this.showLogDetailWrapper.style.display = 'none';
+        this.logTableWrapper.classList.remove('width-60');
+      }
+    });
```
rpl, line 127: All these event handlers could be merged into the `handleEvent` method in this class, and then inside there we may decide if the event was expected and how we should handle it, e.g. and handleEvent would be looking more or less like:
```js
    if (event.type === 'click') {
      const logDetails = event.target.closest('tr')?._log;

      if (logDetails) {
        // Call the method to open the log details in the sidebar.
        return;
      }

      switch (event.target) {
      case this.saveLogBtn:
        // Call the method that will save the logs to file
        break; 
      case this.closeBtn:
        // Call the method that will close the sidebar.
        break;
      default:
        // Throw error for the click event on an unexpected element.
      }
    } else if (event.type === 'change' && event.target == this.logInputFile) {
      // Load the file..      
    } else {
      // throw an error for an unexpected event type
    }
```

src/activitylog/activitylog.html
```diff
@@ -1 +1,46 @@
...
+        <div class="activity-log-container">
+          <div class="log-table">
+            <table>
+              <tr>
+                <th>Extension Id</th>
+                <th>Timestamp</th>
+                <th>API Type</th>
+                <th>API Name</th>
+                <th>View Type</th>
+              </tr>
```
rpl, line 30: Nit, this `<tr>` should be moved into a `<thead>` tag otherwise the table is going to have two `<tbody>` child elements (because the browser creates one automatically to move this `<tr>` tag into it, and then the row created using `insertRow` will be added to the `<tbody>` created automatically, and the `<tbody>` that follows at line 31 will still be there in the DOM but empty).

src/lib/ext-activitylog.js
```diff
@@ -0,0 +1,143 @@
...
+        this.notice.textContent = '';
+        this.notice.classList.remove('failure');
+      }
+    } catch (error) {
+      this.notice.textContent = error.message;
+      this.notice.classList.add('failure');
+    }
+  }
+
+  async saveLogs(logs) {
```
rpl, line 48: Nit, in my opinion this method should be only responsible for saving the logs that are stored in `this.logs` and so it would not need to get the array of logs as a parameter, can just use `this.logs` internally.

src/lib/ext-activitylog.js
```diff
@@ -0,0 +1,143 @@
...
+
+  async getExistingLogs() {
+    const { existingLogs } = await browser.runtime.sendMessage({
+      requestType: 'sendAllLogs',
+      requestTo: 'ext-monitor',
+    });
+    return existingLogs;
+  }
+
+  async loadFile(file) {
```
rpl, line 31: `loadFile` could also avoid to take a file as a parameter, it could instead be responsible also of retrieving the file from `this.logInputFile` internally (in other words the event listener should be able to just call it without any parameter, and everything that should happen could be part of the `loadFile` method itself).

src/lib/ext-activitylog.js
```diff
@@ -0,0 +1,143 @@
...
+      const { requestTo, requestType } = message;
+
+      if (requestTo !== 'activity-log') {
+        return;
+      }
+
+      if (requestType === 'appendLogs') {
+        this.renderNewRow([message.log]);
+      } else {
+        throw new Error('wrong request type found ' + requestType);
```
rpl, line 139: Nit, use a template string here instead.

src/activitylog/activitylog.js
```diff
@@ -1 +1,5 @@
-/* activitylog.js */
+import ActivityLog from '../lib/ext-activitylog.js';
+
+window.addEventListener('DOMContentLoaded', async () => {
```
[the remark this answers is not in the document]
rpl: In most cases may not make any actualy difference, nevertheless technically the target of the DOMContentLoaded event should have the `document` as its actual target and so it may be reasonable to add the listener for this event directly on the `document` element, even if the event listener can be registered on both the `window` and the `document`.

src/lib/ext-activitylog.js
```diff
@@ -0,0 +1,143 @@
...
+    this.saveLogBtn = document.getElementById('saveLogBtn');
+    this.showLogDetailWrapper = document.querySelector(
+      '.show-log-detail-wrapper'
+    );
+    this.showLogDetails = document.getElementById('showLogDetails');
+    this.showLoadedLog = document.getElementById('showLoadedLog');
+    this.logTableWrapper = document.querySelector('.log-table');
+    this.table = document.querySelector('table');
+    this.notice = document.querySelector('.notice');
+    this.closeBtn = document.querySelector('.close');
```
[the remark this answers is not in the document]
rpl: That's a reasonable point, and it may as well be said for ext-popup.js.

In the last meeting we briefly discussed of using the WebComponents API for the classes that are going to manage the DOM elements, but we agreed to defer that to follow ups. 
As part of those follow ups we may also move the ESM modules that are going to export WebComponents classes into a separate lib/web-components/ directory (and tweak the class names a bit, e.g. by adding a `*Element' suffix).

src/lib/ext-activitylog.js
atiqueahmedziad: Currently, I am showing the most recent logs at the top of the table. So, this is inserting a new row after the first row which was the row `th` elements. 
@rpl Do we want to append the most recently logs to the end of the table?

src/activitylog/activitylog.js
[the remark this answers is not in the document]
atiqueahmedziad: Okay, using `document` then :-)

src/lib/ext-activitylog.js
```diff
@@ -0,0 +1,143 @@
...
+  }
+
+  async saveLogs(logs) {
+    try {
+      await save.saveAsJSON(logs);
+
+      if (this.notice.textContent.trim().length) {
+        this.notice.textContent = '';
+        this.notice.classList.remove('failure');
+      }
```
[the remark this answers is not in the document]
atiqueahmedziad: done.

src/lib/ext-activitylog.js
```diff
@@ -0,0 +1,143 @@
...
+    } catch (error) {
+      this.notice.textContent = error.message;
+      this.notice.classList.add('failure');
```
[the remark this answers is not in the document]
atiqueahmedziad: Thanks. Done :-)

src/lib/ext-activitylog.js
```diff
@@ -0,0 +1,143 @@
...
+    }
+  }
+
+  renderSideBar(logIndex) {
+    const logStringfy = JSON.stringify(this.logs[logIndex]);
+    this.showLogDetails.textContent = logStringfy;
+    this.showLogDetailWrapper.style.display = 'block';
+    this.logTableWrapper.classList.add('width-60');
+  }
```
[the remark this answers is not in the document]
atiqueahmedziad: Thanks, done :-)

src/lib/ext-activitylog.js
```diff
@@ -0,0 +1,143 @@
...
+
+  renderNewRow(logs) {
+    for (const log of logs) {
+      const newRow = this.table.insertRow(1);
+      newRow.insertCell(0).textContent = log.id;
+      newRow.insertCell(1).textContent = log.timeStamp;
+      newRow.insertCell(2).textContent = log.type;
+      newRow.insertCell(3).textContent = log.name;
+      newRow.insertCell(4).textContent = log.viewType || 'undefined';
+      newRow.setAttribute('index', this.rowIndex++);
```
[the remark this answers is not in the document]
atiqueahmedziad: Yeah, then we don't need an extra attribute. Done :-)

src/lib/ext-activitylog.js
atiqueahmedziad: For now, I have taken off this part. I will send a follow up pr for load logs.

src/lib/ext-activitylog.js
atiqueahmedziad: Yeah, to avoid the `th` elements, I am currently adding the event listener to table's tbody (where all the logs are represented as `tr`). Is this okay ?

src/lib/ext-activitylog.js
atiqueahmedziad: Thanks for noticing. It's done now :-)

src/activitylog/activitylog.html
atiqueahmedziad: done :-) Sorry I missed this out earlier.

src/lib/ext-activitylog.js
atiqueahmedziad: Yeah, I have updated this method excluding the parameter and using `this.logs` internally, but in the future when we may need to save logs after filtering the logs in the table, it could be helpful if we pass the filtered logs as an argument. That was my thinking of having a parameter here.

src/lib/ext-activitylog.js
atiqueahmedziad: For now, I have taken `load log` part off. In the follow-up pr with load logs, I will take care of this suggestion :-)

src/lib/ext-activitylog.js

src/lib/ext-activitylog.js
```diff
@@ -0,0 +1,120 @@
...
+
+  closeDetailSidebar() {
+    this.showLogDetailWrapper.style.display = 'none';
+    this.logTableWrapper.classList.remove('width-60');
+  }
+
+  async saveLogs() {
+    try {
+      await save.saveAsJSON(this.logs);
+      this.setError(false);
```
atiqueahmedziad, line 115: I put `false` here for readability purpose.

rpl: using false to clear the event makes the expected type signature of the `setError` method a bit odd:

- by passing a string or null/undefined, we can say that the type signature for `setErrot` would be something like `setError(errorMessage: string|void): void` (gets a string or nothing and return nothing)

- by passing an error message `string` or `false`, then `setError` would have a type signature that would look like `setError(errorMessage: string|bool): void` (gets a string or a boolean and return nothing)

In my opinion, `setError(errorMessage: string|bool): void` is a more reasonable type signature for that function.

atiqueahmedziad: Currently, it is in `setError(errorMessage: string|bool): void` format, right ? Like `errorMessage: false` if I am sending `false` as argument. And `errorMessage: string` when I am sending the `error.message` as argument.

src/lib/ext-activitylog.js
```diff
@@ -0,0 +1,122 @@
...
+      switch (event.target) {
+        case this.saveLogBtn:
+          this.saveLogs();
+          break;
+        case this.closeBtn:
+          this.closeDetailSidebar();
+          break;
+        default:
+          throw new Error(
+            `unexpected click event on ${JSON.stringify(event.target)}`
```
Rob--W, line 94: `event.target` is a DOM element, and typically not JSON-serializable. Consider using `event.target.tagName` instead.

src/lib/ext-activitylog.js
```diff
@@ -0,0 +1,122 @@
...
+          );
+      }
+    } else {
+      throw new Error(`wrong event type found ${JSON.stringify(event.type)}`);
```
Rob--W, line 98: `event.type` is already a string without special characters, so `JSON.stringify` is not needed. Use`"${event.type}"` instead.

src/activitylog/activitylog.css
```diff
@@ -1 +1,115 @@
...
+  position: relative;
+}
+
+.show-log-detail-wrapper {
+  position: absolute;
+  top: 0;
+  height: 500px;
+  right: 0;
+  width: 40%;
+  display: none;
```
Rob--W, line 28: It seems that you're using `display: none` with the intention of hiding the wrapper by default, combined with assigning to `.style.display` to toggle its visibility.

Instead of defaulting to `display:none` here, set the `hidden` attribute in HTML, and toggle the `hidden` attribute / property in JavaScript.

- https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/hidden
- https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/hidden

src/lib/ext-activitylog.js
```diff
@@ -0,0 +1,122 @@
...
+      switch (event.target) {
+        case this.saveLogBtn:
+          this.saveLogs();
+          break;
+        case this.closeBtn:
+          this.closeDetailSidebar();
+          break;
+        default:
+          throw new Error(
+            `unexpected click event on ${JSON.stringify(event.target.tagName)}`
```
rpl, line 94: tagName is a string, no need to stringify it.

PR #6, opened 2020-06-29
chore: added extension installation instructions

atiqueahmedziad:
Added extension installation instruction and `web-ext` as devDependency.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,3 +1,35 @@
-# Extension Activity Monitor
+# Extension Activity Monitor [![Build Status](https://travis-ci.org/mozilla/extension-activity-monitor.svg?branch=master)](https://travis-ci.org/mozilla/extension-activity-monitor) [![Coverage Status](https://coveralls.io/repos/github/mozilla/extension-activity-monitor/badge.svg?branch=master)](https://coveralls.io/github/mozilla/extension-activity-monitor?branch=master)
 
-TBD
+This is a privileged Firefox extension that uses the `activityLog` API to monitor the activities of the other installed extensions.
+
+## Installation
+
+### Prerequisite
+
+- Clone / download the respository in your local machine.
+- Install [Firefox Nightly](https://www.mozilla.org/en-US/firefox/all/#product-desktop-nightly) to run this privileged extension.
+- [optional] run `$ npm ci` to install dev dependency to develop a patch or to run the extension using the web-ext dependency.
+
+### Using [web-ext](https://github.com/mozilla/web-ext)
+
+Go to project directory and and run the following command:
+
+```
+$ npm start
+```
+
+_NOTE:_ You may need to run Firefox Nightly once to let the above command to auto-discovery where the Firefox Nightly binary is located in your system.
+Alternatively the Firefox binary location can be provided manually on the command line:
+
+```
+$ web-ext run -f /path/to/firefox-nightly/firefox
+```
+
+### Manual Installation
+
+- Open Firefox Nightly and go to the following URL: `about:config`.
+  - Set `extensions.experiments.enabled` to `true`.
+- Go to the following URL: `about:debugging#/runtime/this-firefox`.
+  - Click on "Load Temporary Add-on" button, then go to `src` directory of your cloned/ downloaded repository and choose `manifest.json` file.
+
+The extension will be loaded temporarily in the browser.
--- a/package.json
+++ b/package.json
@@ -3,6 +3,7 @@
   "version": "1.0.0",
   "description": "Extension Activity Monitor provides more transparency which could help to increase the reliability of abuse reporting and accountability on developers. It is an additional useful tool to aid investigating bugs in the extensions or in the Firefox WebExtensions internals.",
   "scripts": {
+    "start": "web-ext run",
     "test": "jest",
     "test:watch": "jest --watch",
     "test:coverage": "jest --coverage",
@@ -43,6 +44,16 @@
     "eslint-config-prettier": "^6.11.0",
     "eslint-plugin-import": "^2.21.2",
     "jest": "^26.0.1",
-    "prettier": "2.0.5"
+    "prettier": "2.0.5",
+    "web-ext": "^4.3.0"
+  },
+  "webExt": {
+    "sourceDir": "src",
+    "run": {
+      "pref": [
+        "extensions.experiments.enabled=true"
+      ],
+      "firefox": "nightly"
+    }
   }
 }
--- a/package-lock.json
+++ b/package-lock.json
@@ -901,6 +901,16 @@
         "@babel/helper-plugin-utils": "^7.10.1"
       }
     },
+    "@babel/polyfill": {
+      "version": "7.7.0",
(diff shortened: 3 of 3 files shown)


package.json
```diff
@@ -3,6 +3,7 @@
   "version": "1.0.0",
   "description": "Extension Activity Monitor provides more transparency which could help to increase the reliability of abuse reporting and accountability on developers. It is an additional useful tool to aid investigating bugs in the extensions or in the Firefox WebExtensions internals.",
   "scripts": {
+    "start:firefox": "web-ext run -s src --firefox-binary /path/to/firefox-nightly/firefox --pref=\"extensions.experiments.enabled=true\"",
```
Rob--W, line 6: You don't have to manually specify the binary location. If you use `-f nightly` (`-f` is short for `--firefox-binary`), then `web-ext` (or rather its dependency at https://github.com/mozilla-jetpack/node-fx-runner ) will locate the binary. You only need to set a path if it is not in a standard location. But that user-specific location should not be hard-coded in the `package.json` file, because it will likely not work for anyone else.

package.json
```diff
@@ -3,6 +3,7 @@
   "version": "1.0.0",
   "description": "Extension Activity Monitor provides more transparency which could help to increase the reliability of abuse reporting and accountability on developers. It is an additional useful tool to aid investigating bugs in the extensions or in the Firefox WebExtensions internals.",
   "scripts": {
+    "start:firefox": "web-ext run -s src -f nightly --pref=\"extensions.experiments.enabled=true\"",
```
rpl, line 6: @atiqueahmedziad if we do add a "webExt" property to this package.json file, as I did mention in our meeting earlier today ([also documented here in this doc page related to the web-ext cli tool](https://extensionworkshop.com/documentation/develop/getting-started-with-web-ext/#setting-option-defaults-in-a-configuration-file)), then we should be able to:
- reduce this command to the bare minimum (`"start:firefox": "web-ext run"`)
- reduce the comment suggested if the user want to pass other options to just the options that would differ (e.g. `web-ext run --firefox-binary /path/to/your/nightly/firefox`), because  all the other parameters would default to the ones specified in the configuration included in the package.json file.

The configuration to be included in the package.json should look more or less like:
```json
  "devDependencies": {
    ...
  },
  "webExt": {
    "sourceDir": "src",
    "run": {
      "pref": ["extensions.experiments.enabled=true"],
      "firefox": "nightly"
    }
  }
}
```


Given that this extension isn't going to run on anything but Firefox Desktop (at least at the moment), we may call the script just "start" instead of "start:firefox".

package.json
```diff
@@ -43,6 +44,7 @@
     "eslint-config-prettier": "^6.11.0",
     "eslint-plugin-import": "^2.21.2",
     "jest": "^26.0.1",
-    "prettier": "2.0.5"
+    "prettier": "2.0.5",
```
rpl, line 47: It is unrelated to this PR, but I just noticed that prettier is the only dev dependency that is using an "exact dependency" instead of the "^version" used for the other dependencies (where "^version" means "Compatible with version" and would relax the dependency to also include new patch releases, See https://docs.npmjs.com/files/package.json#dependencies for more information about the meaning of the syntax that can be used to specify the version of the dependencies).

We may want to be consistent and use the "^version" also for prettier, no need to rush on it, I mainly wanted to mention it. If we want to change that, I would suggest to do it in a separate pull request.

README.md
```diff
@@ -1,3 +1,34 @@
...
+### Prerequisite
+
+- Clone / download the respository in your local machine.
+- Install [Firefox Nightly](https://www.mozilla.org/en-US/firefox/all/#product-desktop-nightly) to run this privileged extension.
+
+### Using [web-ext](https://github.com/mozilla/web-ext)
+
+- Run `npm run start:firefox` to load the extension temporarily in the Firefox Nightly.
+
+_NOTE_: You need to open Firefox Nightly previously to run the above command. If Firefox Nightly is undetectable or if you have installed it in a different location you may try the following command:
```
rpl, line 16: Nit (just a couple of tweaks, and a shorter command to override just the firefox binary, if we do include the rest of the options in the web-ext configuration as suggested in my previous comment): 

_NOTE_: You may need to run Firefox Nightly once to let the above command to auto-discovery where the Firefox Nightly binary is located in your system.
Alternatively the Firefox binary location can be provided manually on the command line:
```
$ web-ext run -f /path/to/firefox-nightly/firefox
```

README.md
```diff
@@ -1,3 +1,34 @@
-# Extension Activity Monitor
+# Extension Activity Monitor [![Build Status](https://travis-ci.org/mozilla/extension-activity-monitor.svg?branch=master)](https://travis-ci.org/mozilla/extension-activity-monitor) [![Coverage Status](https://coveralls.io/repos/github/mozilla/extension-activity-monitor/badge.svg?branch=master)](https://coveralls.io/github/mozilla/extension-activity-monitor?branch=master)
 
-TBD
+This is a privileged extension for Firefox that monitors the activities of other installed extensions. The extension uses a privileged API called `activityLog` API to retrieve logs from other installed extensions.
```
rpl, line 3: Nit, we could merge the two phrases I guess: "This is a privileged Firefox extensions that uses the `activityLog` API to monitor the activities of the other installed extensions."

README.md
```diff
@@ -1,3 +1,34 @@
...
+
+## Installation
+
+### Prerequisite
+
+- Clone / download the respository in your local machine.
+- Install [Firefox Nightly](https://www.mozilla.org/en-US/firefox/all/#product-desktop-nightly) to run this privileged extension.
```
rpl, line 10: Nit:
- Clone ...
- Install ...
- [optional] `npm ci`, to install dev dependency to develop a patch or to run the extension using the web-ext dependency

README.md
```diff
@@ -1,3 +1,34 @@
...
+_NOTE_: You need to open Firefox Nightly previously to run the above command. If Firefox Nightly is undetectable or if you have installed it in a different location you may try the following command:
+
+```
+ web-ext run -s src -f /path/to/firefox-nightly/firefox --pref="extensions.experiments.enabled=true"
+```
+
+### Manual Installation
+
+You can follow any of the following two ways:
+
```
rpl, line 25: For the Manual Installation it may be worth to mention explicitly that "extensions.experiments.enabled" should be set to true from about:config.

package.json
atiqueahmedziad: Sorry, I should have asked about it since I was confused with setting up `pref`. Thanks for explaining. Added it.

package.json
atiqueahmedziad: Yeah, will send a pr :-)

README.md
atiqueahmedziad: Thanks for the suggestion. Added it.

README.md
atiqueahmedziad: Thanks for the suggestion. Description updated.

README.md
atiqueahmedziad: Added it :-)

README.md
atiqueahmedziad: I have updated this part to a single way for Manual Installation and mention the `extensions.experiments.enabled` part. Is this okay ?

PR #8, opened 2020-06-30
test: disallow unsanitized code with eslint check

atiqueahmedziad:
As discussed, we are disallowing unsanitized code using the `eslint-plugin-no-unsanitized`.

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -1,5 +1,10 @@
 {
-  "extends": ["eslint:recommended", "plugin:import/errors", "prettier"],
+  "extends": [
+    "eslint:recommended",
+    "plugin:import/errors",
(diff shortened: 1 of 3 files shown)


atiqueahmedziad:
It already found some problems :-p Should we skip checking unsanitized code in tests?

rpl:
> It already found some problems :-p Should we skip checking unsanitized code in tests?

@atiqueahmedziad  That's fine, test code is not really exposing that to arbitrary content passed by a third party.
Let's just disable the `"no-unsanitized/property"` rule in the test files

(there are a couple of different ways to achieve that, but at the moment there is only one test file that is triggering it and so we could just disable the rule in that single file using an eslint configuration comment, see https://eslint.org/docs/user-guide/configuring#using-configuration-comments).

PR #9, opened 2020-07-08
feat: table view webcomponent is added and mvc pattern is followed

atiqueahmedziad:
Table view webcomponent is added and mvc pattern is followed


src/lib/web-component/log-table-element.js
```diff
@@ -0,0 +1,68 @@
+class LogTable extends HTMLElement {
+  constructor() {
+    const self = super();
```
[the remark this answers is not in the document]
atiqueahmedziad: Done :-)

src/lib/web-component/log-table-element.js
```diff
@@ -0,0 +1,68 @@
...
+    const logTableInstance = document.importNode(
+      this.logTableTemplate.content,
+      true
+    );
+    this.tableBody = logTableInstance.querySelector('tbody');
+    this.logTableWrapper = logTableInstance.querySelector('.log-table-wrapper');
+
+    shadow.appendChild(style);
+    shadow.appendChild(logTableInstance);
+    shadow.appendChild(this.detailSideBar);
```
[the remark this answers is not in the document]
atiqueahmedziad: For now, `detailSidebar` codes are dropped.

src/lib/web-component/log-table-element.js
```diff
@@ -0,0 +1,66 @@
+class LogTable extends HTMLElement {
+  constructor() {
+    const self = super();
+    const shadow = this.attachShadow({ mode: 'open' });
+
+    this.detailSideBar = self.querySelector('detail-sidebar');
+    this.closeBtn = this.detailSideBar.shadowRoot.querySelector('.close');
+
+    const style = document.createElement('style');
+    style.textContent = `@import '/activitylog/activitylog.css'`;
```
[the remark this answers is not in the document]
atiqueahmedziad: done. Thanks :-)

src/activitylog/activitylog.html
```diff
@@ -15,24 +17,23 @@ <h1>Extension Activity Monitor</h1>
...
-              </thead>
-              <tbody></tbody>
-            </table>
-          </div>
-          <div class="show-log-detail-wrapper" hidden>
-            <div id="showLogDetails"></div>
-            <div class="close"></div>
-          </div>
+          <log-table>
+            <template id="log-table-template">
```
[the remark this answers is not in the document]
atiqueahmedziad: done :-)

src/lib/web-component/log-view/log-view-element.js
```diff
@@ -0,0 +1,70 @@
...
+  constructor() {
+    super();
+    const shadow = this.attachShadow({ mode: 'open' });
+
+    this.logDetailWrapper = document.querySelector('.log-detail-wrapper');
+    this.logDetails = document.querySelector('#logDetails');
+    this.closeBtn = document.querySelector('.close');
+
+    this.logTableTemplate = document.querySelector('.log-table-template');
+    this.logTableWrapper = document.querySelector('.log-table-wrapper');
```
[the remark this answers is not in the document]
atiqueahmedziad: I made the `<log-view>` self-contained now. I was thinking of using `<slot>` later when I make the sidebar as separate component. Is that okay ?

src/lib/ext-activitylog.js
```diff
@@ -1,30 +1,47 @@
...
+    this.view = view;
+
+    this.init();
+  }
 
-    this.saveLogBtn.addEventListener('click', this);
-    this.closeBtn.addEventListener('click', this);
-    this.tableBody.addEventListener('click', this);
+  async init() {
+    this.view.saveLogBtn.addEventListener('click', this);
```
Rob--W, line 44: To decouple the view's DOM from this high-level View class, subscribe for a custom event on the view. In the view, you can use the `CustomEvent` to create a new event and fire it. See https://developer.mozilla.org/en-US/docs/Web/Guide/Events/Creating_and_triggering_events

src/activitylog/activitylog.html
```diff
@@ -14,27 +18,38 @@ <h1>Extension Activity Monitor</h1>
...
-          </div>
-          <div class="show-log-detail-wrapper" hidden>
-            <div id="showLogDetails"></div>
-            <div class="close"></div>
-          </div>
-        </div>
+        <log-view></log-view>
       </div>
     </div>
+    <template class="log-table-template">
```
Rob--W, line 24: There is only one template in the document, use the `id` attribute instead.

(also at the next template below).

And add a blank line before the `<template ...>` element to slightly improve readability.

src/activitylog/activitylog.html
```diff
@@ -14,27 +18,38 @@ <h1>Extension Activity Monitor</h1>
...
+              <th>Timestamp</th>
+              <th>API Type</th>
+              <th>API Name</th>
+              <th>View Type</th>
+            </thead>
+            <tbody></tbody>
+          </table>
+        </div>
+        <div class="log-detail-wrapper" hidden>
+          <div id="logDetails"></div>
```
Rob--W, line 40: Use `class` instead of `id` when there may be more than one instance of an element. Anything in a custom element could in theory be used more than once, so use `class="logDetails"` instead.

src/lib/web-component/log-view/log-view-element.js
```diff
@@ -0,0 +1,87 @@
...
+
+  openDetailSidebar(logDetails) {
+    const logString = JSON.stringify(logDetails);
+    this.logDetails.textContent = logString;
+    this.logTableWrapper.classList.add('width-60');
+    this.logDetailWrapper.removeAttribute('hidden');
+  }
+
+  closeDetailSidebar() {
+    this.logDetailWrapper.setAttribute('hidden', true);
```
Rob--W, line 56: This works, but not for the reason that you expect.

`setAttribute` can only assign string values as attribute values, so
`.setAttribute('hidden', true);` is equivalent to `.setAttribute('hidden', 'true');`

`.setAttribute('hidden', false);` would be equivalent to `.setAttribute('hidden', 'false');`, which does NOT unhide the element.

Use `.hidden = true;` instead, which is equivalent to `.setAttribute('hidden', '');`
(and `.hidden = false;` is equivalent to `.removeAttribute('hidden');`. )

See https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/hidden
and in general https://stackoverflow.com/questions/4139786/what-does-it-mean-in-html-5-when-an-attribute-is-a-boolean-attribute

src/lib/web-component/log-view/log-view-element.js
```diff
@@ -0,0 +1,87 @@
...
+    if (event.type === 'click') {
+      const logDetails = event.target.closest('tr')?._log;
+
+      if (logDetails) {
+        this.openDetailSidebar(logDetails);
+        return;
+      }
+
+      if (event.target === this.closeBtn) {
+        this.logTableWrapper.classList.remove('width-60');
```
Rob--W, line 70: Remove this. This is already done in `closeDetailSidebar`.

src/activitylog/activitylog.html
```diff
@@ -14,27 +18,38 @@ <h1>Extension Activity Monitor</h1>
...
-          <div class="show-log-detail-wrapper" hidden>
-            <div id="showLogDetails"></div>
-            <div class="close"></div>
-          </div>
-        </div>
+        <log-view></log-view>
       </div>
     </div>
+    <template class="log-table-template">
+      <link rel="stylesheet" href="/lib/web-component/log-view/log-view.css" />
```
Rob--W, line 25: Remove `/` from `/>`. This is not XHTML, the `/` serves no purpose.

src/lib/web-component/log-view/log-view-element.js
```diff
@@ -0,0 +1,87 @@
...
+    shadow.appendChild(logTableInstance);
+  }
+
+  addNewRows(logs) {
+    for (const log of logs) {
+      const logTableRowInstance = document.importNode(
+        this.logTableRowTemplate.content,
+        true
+      );
+      logTableRowInstance.querySelector('tr')._log = log;
```
Rob--W, line 34: Ideally you would assign directly to `logTableRowInstance`. Assuming that the template contains one element only, you could achieve the desired goal by using `this.logTableRowTemplate.content.firstElementChild`.

And instead of reading that property so often here, you could directly read that from the constructor, so you can simply do `const logTableRowInstance = this.logTableRowTemplate.cloneNode(true);` here. (cloneNode and `document.importNode` are equivalent, but cloneNode is shorter).

src/lib/web-component/log-view/log-view-element.js
```diff
@@ -0,0 +1,87 @@
...
+
+      logTableRowInstance.querySelector('#id').textContent = log.id;
+      logTableRowInstance.querySelector('#timestamp').textContent =
+        log.timeStamp;
+      logTableRowInstance.querySelector('#apiType').textContent = log.type;
+      logTableRowInstance.querySelector('#name').textContent = log.name;
+      logTableRowInstance.querySelector('#viewType').textContent =
+        log.viewType || 'undefined';
+
+      this.tableBody.appendChild(logTableRowInstance);
```
Rob--W, line 44: In general, if you are going to append lots of DOM elements, use `document.createDocumentFragment` first, append the elements there, and after the end of the loop, append the document fragment to the desired destination. This results in a better performance.

src/lib/web-component/log-view/log-view.css
```diff
@@ -0,0 +1,85 @@
...
+  display: flex;
+  width: 100%;
+}
+
+tbody tr:nth-of-type(odd) {
+  background: #eeeeee;
+}
+
+tbody tr {
+  display: flex;
```
Rob--W, line 75: Is it really necessary to change from `display: table-row` (default) to `display: flex`?

src/activitylog/activitylog.html
atiqueahmedziad: It is done by prettier. I can tell prettier to ignore this if you want.
Note: I faced this issue earlier when Luca asked me to [remove `/` self-closing tag meta tags](https://github.com/mozilla/extension-activity-monitor/pull/3#discussion_r443464519). But [he suggested to not fight with prettier for it.](https://github.com/mozilla/extension-activity-monitor/pull/3#discussion_r444405359)

Rob--W: Never mind then. Prettier is wrong, but it's not worth the effort.

src/lib/ext-activitylog.js

src/activitylog/activitylog.html
atiqueahmedziad: done :-)

src/activitylog/activitylog.html
atiqueahmedziad: done :-)

src/lib/web-component/log-view/log-view-element.js
atiqueahmedziad: done. Thanks :-)

src/lib/web-component/log-view/log-view-element.js
atiqueahmedziad: ahh sorry for missed it earlier :(

src/lib/web-component/log-view/log-view-element.js
atiqueahmedziad: done :-)

src/lib/web-component/log-view/log-view-element.js
atiqueahmedziad: Thanks, done :-)

src/lib/web-component/log-view/log-view.css
atiqueahmedziad: Sorry forget to change this, done now. :-)

src/activitylog/activitylog.html
```diff
@@ -14,27 +18,40 @@ <h1>Extension Activity Monitor</h1>
...
+          </table>
+        </div>
+        <div class="log-detail-wrapper" hidden>
+          <div class="log-details"></div>
+          <div class="close"></div>
+        </div>
+      </div>
+    </template>
+
+    <template class="log-table-row-template">
```
Rob--W, line 47: This should also use an id attribute.

src/activitylog/activitylog.html
```diff
@@ -14,27 +18,40 @@ <h1>Extension Activity Monitor</h1>
...
+      <tr>
+        <td id="id"></td>
```
Rob--W, line 49: ... and this should use a "class" attribute.

src/lib/web-component/log-view/log-view-element.js
```diff
@@ -0,0 +1,85 @@
...
+    this.tableBody = logTableInstance.querySelector('tbody');
+
+    shadow.appendChild(logTableInstance);
+  }
+
+  addNewRows(logs) {
+    const rowsFragment = document.createDocumentFragment();
+
+    for (const log of logs) {
+      const logTableRowInstance = this.logTableRowTemplate.content.cloneNode(
```
Rob--W, line 28: I suggest to store the firstElementChild in th constructor, so you don't have to clone the superfluous whitespace. This code is relatively hot (used often), so it'll be good to minimize the amount of redundant calls.

Furthermore, by storing it in the constructor, it will be more obvious that there is only one (root) node in the template.

src/lib/web-component/log-view/log-view-element.js
```diff
@@ -4,10 +4,11 @@ class LogView extends HTMLElement {
 
     const shadow = this.attachShadow({ mode: 'open' });
 
-    this.logTableTemplate = document.querySelector('#log-table-template');
-    this.logTableRowTemplate = document.querySelector(
-      '.log-table-row-template'
-    );
+    this.logTableRow = document
+      .querySelector('#logTableRowTemplate')
+      .content.firstElementChild.cloneNode(true);
```
Rob--W, line 9: `cloneNode(true);` isn't needed. You're going to `cloneNode(true);` this later anyway.

Rob--W:
LGTM!

PR #12, opened 2020-07-20
feat: clear logs from activitylog page

atiqueahmedziad:


src/lib/ext-activitylog.js
```diff
@@ -24,11 +28,16 @@ class Model {
       this.filter[logKey].splice(logKeyIndex, 1);
     }
   }
+
+  clearLogs() {
+    this.init();
```
Rob--W, line 33: I would just clear the logs at first, and not the filters.
If you really wish to include filter reset functionality (I don't think that it's necessary, because a user can reload the page to achieve the same result), this logic may work:

- Click once - clear logs.
- If filters are set - update button to clear filters.
- If filters are updated or logs are updated, show "clear logs" on button again.

In terms of usability, clearing filters upon clearing the log would be comparable to losing all breakpoints when you reload a page that you're debugging.

atiqueahmedziad: @Rob--W Instead, should we have a separate `Clear Filter` button, that will -
- Removes all the filter checkboxes, when there are no logs in the table view (most likely after clicking clear logs button)
- Enables all the checkboxes (clear the filter object in `model`, i.e. no filters are being applied in the table view), when there are logs in the table view.

How does this sound?

atiqueahmedziad: @Rob--W I will do the `Clear Filter` feature in a follow-up pr later (I will write it in an issue as future follow-ups, once we decide how we want to do it). 
I am focusing on **clearing logs only** in this pr. I have updated the pr. :-)

PR #13, opened 2020-07-21
feat: filter option with API name

atiqueahmedziad:


atiqueahmedziad:
Yeah, since I was getting API names mostly and that sound more meaningful. But I think I can re-name the variable as `nameFilter` instead of `apiNameFilter`. Also as checkbox's label, I can put `Select script or API name to filter logs`. How does this sound?

Rob--W:
This sounds like a suboptimal UI. When filtering by API name, I would expect to only match API names.

The current table and UI is guided by the internal representation of the activity logs. I suggest to deviate from that representation when a more optimal user experience is available.

It probably makes sense to group content scripts with viewType (viewType is only set for extension pages, content scripts can only run in non-extension pages).

atiqueahmedziad:
> The current table and UI is guided by the internal representation of the activity logs. I suggest to deviate from that representation when a more optimal user experience is available.

Yes, I can do that by showing the content script url, under `view type` instead of `API Name`. So, the API Name column for any `content_script` will be `undefined` then? 
Also, do you want me to do it in a follow-up or in this pr?

> It probably makes sense to group content scripts with viewType (viewType is only set for extension pages, content scripts can only run in non-extension pages).

I was also thinking, is it better to fix it from API's endpoint since this sounds reasonable from API's perspective too?

Rob--W:
> Yes, I can do that by showing the content script url, under `view type` instead of `API Name`

It's actually not exactly `viewType` either. 
In the table, the "API name" + "viewType" columns could be combined (with `td.colSpan = 2`) and the script be rendered there.
In the options dropdown boxes, the script URL doesn't fit in any of the existing options. A separate filter box would make more sense.

> Also, do you want me to do it in a follow-up or in this pr?

The minimum in this PR is to not put the scripts themselves in the "API name" filter. You could render a new category "other" (instead of `undefined`) that is the catch-all for anything not represented by the checklist.

> I was also thinking, is it better to fix it from API's endpoint since this sounds reasonable from API's perspective too?

`viewType` is an enum with limited values, so I don't think that it makes sense to also add the script URL there.

atiqueahmedziad:
> The minimum in this PR is to not put the scripts themselves in the "API name" filter. You could render a new category "other" (instead of `undefined`) that is the catch-all for anything not represented by the checklist.

@Rob--W Did you mean to make another filter option named `other` where there will be content script urls (for now) as checkboxes?

src/lib/web-component/filter-option/filter-option-element.js
```diff
@@ -26,6 +28,10 @@ class FilterOption extends HTMLElement {
 
   updateFilterCheckboxes(logs) {
     for (const log of logs) {
+      if (this.filterKey === 'name' && log.type === 'content_script') {
```
Rob--W, line 31: This doesn't belong here. You should control this at the input, e.g. by using `logs.filter(log => ...)` to exclude some entries.

atiqueahmedziad:
> In the table, the "API name" + "viewType" columns could be combined (with td.colSpan = 2) and the script be rendered there.

This is done now. :-)

> > Also, do you want me to do it in a follow-up or in this pr?
> 
> The minimum in this PR is to not put the scripts themselves in the "API name" filter. You could render a new category "other" (instead of `undefined`) that is the catch-all for anything not represented by the checklist.

In the current version, I have placed the content script URLs under a new filter option named "Other". Now, the API name filter will only contain API names. 
Also, 
I will file an issue, to move the `undefined` checkbox from viewType to "Other" filter option and work on it in any future pr.

src/activitylog/activitylog.html
```diff
@@ -30,13 +30,19 @@ <h2>Activity Logs</h2>
...
+            Select view type to filter
           </filter-option>
           <filter-option filter-key="type">
-            Select api type to filter logs
+            Select api type to filter
+          </filter-option>
+          <filter-option filter-key="name">
+            Select api name to filter
+          </filter-option>
+          <filter-option filter-key="other">
```
Rob--W, line 44: This is not what I meant.

> Did you mean to make another filter option named other where there will be content script urls (for now) as checkboxes?

We've misunderstood each other. I thought that you suggested to add an option with value "other" in the existing `<filter-option filter-key="name">`.

I meant this:

```
<filter-option filter-key="name">
[x] tabs.create
[x] tabs.update
[x] other <-- New option here that matches content script logs.
</filter-option>
```


... where `other` covers everything else that doesn't have an API name. Or, since the only type of logs are content scripts, call it something like "Content scripts".

atiqueahmedziad: @Rob--W we already get a checkbox named `content_script` inside API Type `filter-option` when any content script is found. Do we want a checkbox with named  "Content scripts" under API Name `filter-option`  too ?

Rob--W: Currently your filter design assumes that the set of options is closed. That is, every log entry matches one of the given options, and every option matches at least one log.

If the UI is clear* that "API name" only matches log entries related to API events (this is indeed reasonable), then you could indeed choose to not add an "other" or "content scripts" checkbox at all.

\* one stupidly simple way to achieve this is by adding a tooltip that explains to which log types the "API name" filter applies.

atiqueahmedziad:
@Rob--W I have updated the PR hiding the content script urls from "API Name" checkboxes and added a tooltip on "API Name" filter option.

src/lib/ext-activitylog.js
```diff
@@ -51,6 +54,12 @@ class Model {
     return this.filter.type.has(logType);
   }
 
+  matchFilterApiName(name) {
+    return (
+      this.filter.name.has(name) || this.filter.contentScriptUrls.has(name)
```
Rob--W, line 59: Don't save `contentScriptUrls` - it is not used elsewhere.

When a filter is not applicable, just return true. You should also do this in your timestamp patch - when there is no timestamp filter don't check the filter at all.

```
if (type is content script) {
  return true;
}
return match normal filter otherwise;
```

src/lib/web-component/log-view/log-view-element.js
```diff
@@ -42,9 +42,20 @@ export class LogView extends HTMLElement {
...
-        log.viewType || 'undefined';
+
+      if (log.type === 'content_script') {
+        const contentScriptTD = logTableRowInstance.querySelector('.name');
+        contentScriptTD.textContent = log.name;
+        contentScriptTD.setAttribute('colspan', '2');
+        // view type is undefined for log.type = content_script
+        logTableRowInstance
+          .querySelector('.view-type')
+          .setAttribute('hidden', true);
```
Rob--W, line 53: `.hidden = true;`

src/lib/web-component/log-view/log-view-element.js
```diff
@@ -42,9 +42,20 @@ export class LogView extends HTMLElement {
...
         log.timeStamp;
       logTableRowInstance.querySelector('.api-type').textContent = log.type;
-      logTableRowInstance.querySelector('.name').textContent = log.name;
-      logTableRowInstance.querySelector('.view-type').textContent =
-        log.viewType || 'undefined';
+
+      if (log.type === 'content_script') {
+        const contentScriptTD = logTableRowInstance.querySelector('.name');
+        contentScriptTD.textContent = log.name;
+        contentScriptTD.setAttribute('colspan', '2');
```
Rob--W, line 49: `.colSpan = 2;`

src/activitylog/activitylog.html
```diff
@@ -30,13 +30,19 @@ <h2>Activity Logs</h2>
...
-            Select view type to filter logs
+            Select view type to filter
           </filter-option>
           <filter-option filter-key="type">
-            Select api type to filter logs
+            Select api type to filter
+          </filter-option>
+          <filter-option
+            filter-key="name"
+            title="It only matches log entries related to API events"
```
Rob--W, line 43: "This filter only applied to log entries related to API events."

src/lib/ext-activitylog.js
```diff
@@ -176,6 +182,23 @@ class Controller {
...
     this.view.addTableRows(logs);
+    this.updateFilterCheckboxes(logs);
+  }
+
+  updateFilterCheckboxes(logs) {
+    const apiNameLogs = logs.filter((log) => {
+      if (log.type !== 'content_script') {
+        return true;
+      } else {
+        //log.name contains content script url
```
Rob--W, line 193: It's good that you put a comment here, because otherwise the logic is not as obvious.

When you remove the `.contentScriptUrls` line  (as I suggested elsewhere), please keep the comment, so that future readers know why content script logs are filtered.

```
// log.name contains the script URL instead of the API name for content scripts.
```

src/activitylog/activitylog.html
```diff
@@ -38,6 +38,12 @@ <h2>Activity Logs</h2>
           <filter-option filter-key="type">
             Select api type to filter logs
           </filter-option>
+          <filter-option
+            filter-key="name"
+            title="This filter only applied to log entries related to API events."
```
Rob--W, line 43: applies*, sorry, I made a typo before.

src/lib/ext-activitylog.js
```diff
@@ -33,26 +35,33 @@ class Model {
...
-  matchFilterId(logId) {
-    return this.filter.id.has(logId);
+  matchFilterId(id) {
+    return this.filter.id.has(id);
   }
 
-  matchFilterViewType(logViewType) {
-    return this.filter.viewType.has(logViewType);
+  matchFilterViewType({ viewType, type }) {
+    return type === 'content_script'
```
Rob--W, line 50: I suggest to use the following logic instead of a ternary.

```
if content script
  // Comment explaining why there is an early return.
  Return true
Return normal check
```

It is more readable, because it is more obvious that content scripts are excluded, together with the reason.

PR #14, opened 2020-07-21
feat: substring filtering option is added

atiqueahmedziad:


src/lib/ext-activitylog.js
```diff
@@ -165,6 +169,23 @@ class Controller {
...
+  onKeywordChange(newKeyword) {
+    this.model.setFilterKeyword(newKeyword);
+    this.view.setLogFilter((log) => this.isFilterMatched(log));
+  }
+
+  isFilterMatched(log) {
+    for (const key of Object.keys(this.model.filter)) {
+      if (this.model.filter[key].includes(log[key])) {
+        return true;
+      } else if (key === 'keyword') {
```
Rob--W, line 182: So now you're looping over all (fixed set of) keys of the filter, and doing conditionals on some keys.

This is not really effective, nor readable. I suggest to put the filter logic in the model, and unroll the loop (i.e. explicitly define the filter logic for every filter key).

src/lib/web-component/filter-keyword/filter-keyword-element.js
```diff
@@ -0,0 +1,38 @@
...
+    const filterContainer = document
+      .querySelector('#filterKeywordTemplate')
+      .content.cloneNode(true);
+
+    this.inputBox = filterContainer.querySelector('input[name="keyword"]');
+
+    shadow.appendChild(filterContainer);
+  }
+
+  connectedCallback() {
```
Rob--W, line 15: Where is the disconnectedCallback?

atiqueahmedziad: Added now.

src/lib/ext-activitylog.js

src/lib/ext-activitylog.js
```diff
@@ -24,6 +25,36 @@ class Model {
...
+    return this.filter.viewType.includes(logViewType);
+  }
+
+  matchFilterType(logType) {
+    return this.filter.type.includes(logType);
+  }
+
+  matchFilterKeyword(logData) {
+    const logDataStr = JSON.stringify(logData);
+    return !logDataStr.includes(this.filter.keyword);
```
Rob--W, line 56: This logic looks inverted/incorrect.

Now, with all filtering logic moved to the model, it should be easier to write unit tests to verify that the filtering logic works as expected. Can you add such unit tests?

src/lib/web-component/filter-keyword/filter-keyword-element.js
```diff
@@ -0,0 +1,41 @@
...
+    this.inputBoxListener = this.delay(
+      (event) => this.onKeywordSubmit(event.originalTarget.value),
```
Rob--W, line 17: Use `event.target` instead of the non-standard originalTarget.

src/lib/ext-activitylog.js
atiqueahmedziad: This is inverted since I have to check from the opposite direction compared to other checkings above. 
Yeah, I can add test for filtering logic.

Rob--W: Why do you want to check the opposite direction?

The log should be shown if the filter matched.

A potential issue of your current logic is that you're using `||`, i.e. showing logs if any of the filters matched. I believe that you should change it to `&&`, show if all filters match.

atiqueahmedziad: It's actually hiding logs when any of the filters match (since I am returning the boolean for setting [`row.hidden`](https://github.com/mozilla/extension-activity-monitor/blob/master/src/lib/web-component/log-view/log-view-element.js#L30) property. Here, I am using the filter object to collect the items which I want to get rid off. 

For example, if the `viewType` key in filter object has value `popup` in it, that means, I am getting rid of the logs whose review type is `popup`.

If I use `&&`, I will also have to invert the check [here](https://github.com/mozilla/extension-activity-monitor/blob/master/src/lib/web-component/log-view/log-view-element.js#L30). Also, currently I am not writing into filter object initially, but in this alternative approach, I will have to store all the filter labels into the filter object and remove them when user unchecks the checkboxes. Should I do that?

Rob--W: Logis should be `.hidden = !filter(log)`

`filter`'s contract is usually "true if the element should be shown", for example with Array's filter: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/filter

src/lib/web-component/filter-keyword/filter-keyword-element.js
atiqueahmedziad: done.

src/lib/ext-activitylog.js
```diff
@@ -7,11 +7,23 @@ class Model {
...
 
   addNewLogs(logs) {
-    this.logs.push(...logs);
+    for (const log of logs) {
+      this.logs.push(log);
+      this.updateFilter(log);
+    }
+  }
+
+  updateFilter(log) {
```
Rob--W, line 21: This logic can't be right. This would re-add a disabled filter whenever a new log entry is added.

atiqueahmedziad: updated :-)

atiqueahmedziad: @Rob--W Why are we doing the same thing (which brings the same result) in an alternate way that involves more code? 
Also, in terms of performance in the current approach, I imagine that we are monitoring a lot of extensions and for that, we are storing all extensions id, type, viewtype... in the model, and comparing filter matching with `isFilterMatched` method involves more time than it would in the previous solution (where `.hidden = filter(log)`).

Rob--W: > @Rob--W Why are we doing the same thing (which brings the same result) in an alternate way that involves more code?

What part of the alternative is required and more code according to you?

In my view, the `addFilter` and `removeFilter` functions are not necessary. When they were added, it was to support the identity filtering method, but it would never work for the substring filtering method. A "filterchange" event that shares the current value of the filter (i.e. the full filter opposed to the addition or removal) would work for the checkbox and keyword input. There is no need for several more events.

I'm kind of surprised to see the introduction of the lot of events here, and am glad that you are questioning the need for them.

> Also, in terms of performance in the current approach, I imagine that we are monitoring a lot of extensions and for that, we are storing all extensions id, type, viewtype... in the model, and comparing filter matching with `isFilterMatched` method involves more time than it would in the previous solution (where `.hidden = filter(log)`).

I don't understand this claim. Why do you believe that there is a significant performance difference between the two?

src/lib/ext-activitylog.js
```diff
@@ -160,15 +198,28 @@ class Controller {
...
+    this.model.setFilterKeyword(newKeyword);
+    this.view.setLogFilter((log) => this.isFilterMatched(log));
+  }
+
+  isFilterMatched(log) {
+    return this.model.matchLogWithFilterObj(log);
+  }
+
+  onFilterUpdate(filterObject) {
+    this.model.addFilter(filterObject);
```
Rob--W, line 217: In isolation, this method seems to be missing a notification to re-apply filtering.

When I look up the caller, it turns out that this filter is only added when a new log entry is added (visible by default).

To clarify that, it would be more useful to reflect that in the method name (e.g. `onNewFilterOption`) and add a comment to explain why it is not necessary to trigger the log updater.

src/lib/ext-activitylog.js
atiqueahmedziad: > > @Rob--W Why are we doing the same thing (which brings the same result) in an alternate way that involves more code?
> 
> What part of the alternative is required and more code according to you?

In previous logic, I didn't need to setup a communication (which I am doing currently [here](https://github.com/mozilla/extension-activity-monitor/pull/14/files#diff-9f25b3c1ffafaec3c4d02ac711c6392aR33-R37) between `filter-option-element` and model to update the model when a new checkbox is added. 

> In my view, the `addFilter` and `removeFilter` functions are not necessary. When they were added, it was to support the identity filtering method, but it would never work for the substring filtering method. A "filterchange" event that shares the current value of the filter (i.e. the full filter opposed to the addition or removal) would work for the checkbox and keyword input. There is no need for several more events.

If I remove `addFilter` and `removeFilter` functions, how do I modify the model's filter object? Modification to filter object in model is required since it says if any of the future logs should hidden or not. 

> > Also, in terms of performance in the current approach, I imagine that we are monitoring a lot of extensions and for that, we are storing all extensions id, type, viewtype... in the model, and comparing filter matching with `isFilterMatched` method involves more time than it would in the previous solution (where `.hidden = filter(log)`).
> 
> I don't understand this claim. Why do you believe that there is a significant performance difference between the two?

There might not be significant performance difference is practical but I was thinking, in the current approach where we save all the ids, viewTypes, type, api names and searching inside them. But previously logic, the filter object would contain only ids, viewTypes, type, api names of those which we want to be hidden. So, there would be less items (unless anyone filters out everything) in the filter object compared to current approach, and searching inside them would be faster for that?

src/lib/ext-activitylog.js
atiqueahmedziad: Should I keep this method ? I will have to keep the `filterupdate` event then. I can also do the same thing with `filterchange` event. What do you suggest ?

src/lib/ext-activitylog.js
Rob--W: > > > @Rob--W Why are we doing the same thing (which brings the same result) in an alternate way that involves more code?
> > 
> > 
> > What part of the alternative is required and more code according to you?
> 
> In previous logic, I didn't need to setup a communication (which I am doing currently [here](https://github.com/mozilla/extension-activity-monitor/pull/14/files#diff-9f25b3c1ffafaec3c4d02ac711c6392aR33-R37) between `filter-option-element` and model to update the model when a new checkbox is added.

This brings up an interesting point: on what is expected to happen when a new log entry is added. When the user isn't filtering anything, the expected result is that the option should be checked by default (this is what you have in mind). But there are use cases where the user is looking for specific API calls, and in that case it would be preferred that new logs that don't match the existing criteria are hidden by default. The two different use cases can be supported by adding a checkbox `[ ] Select all by default` to change whether new options should be checked (or unchecked) by default. This can be done in a follow-up (consider filing an issue in this repo if you want to work on it later).

> > In my view, the `addFilter` and `removeFilter` functions are not necessary. When they were added, it was to support the identity filtering method, but it would never work for the substring filtering method. A "filterchange" event that shares the current value of the filter (i.e. the full filter opposed to the addition or removal) would work for the checkbox and keyword input. There is no need for several more events.
> 
> If I remove `addFilter` and `removeFilter` functions, how do I modify the model's filter object? Modification to filter object in model is required since it says if any of the future logs should be hidden or not.

The filter-option-element has a complete picture of all relevant checkboxes. It could simply send all relevant options to the model (so the model won't have to do `indexOf` + `splice` for example). The controller then doesn't need to know about the format of the filter any more, every `filterchange` event would then replace the filter.

> > > Also, in terms of performance in the current approach, I imagine that we are monitoring a lot of extensions and for that, we are storing all extensions id, type, viewtype... in the model, and comparing filter matching with `isFilterMatched` method involves more time than it would in the previous solution (where `.hidden = filter(log)`).
> > 
> > 
> > I don't understand this claim. Why do you believe that there is a significant performance difference between the two?
> 
> There might not be significant performance difference is practical but I was thinking, in the current approach where we save all the ids, viewTypes, type, api names and searching inside them. But previously logic, the filter object would contain only ids, viewTypes, type, api names of those which we want to be hidden. So, there would be less items (unless anyone filters out everything) in the filter object compared to current approach, and searching inside them would be faster for that?

This is indeed valid for the use case of "show all logs except for X", but doesn't work for "show logs matching Y".
At this stage, it's important to get the prototype working first, optimizations can be done in a follow-up, once it's clear what you need to optimize.

Note: use `Set` instead of `Array` for faster lookups.

src/lib/ext-activitylog.js
```diff
@@ -4,9 +4,9 @@ class Model {
   constructor() {
     this.logs = [];
     this.filter = {
-      id: [],
-      viewType: [],
-      type: [],
+      id: new Set([]),
```
Rob--W, line 7: `new Set()` is sufficient. No need to pass an empty array. Let's fix this in the whole patch.

src/lib/ext-activitylog.js
```diff
@@ -15,19 +15,8 @@ class Model {
...
-    const logKeyIndex = this.filter[logKey].indexOf(valueEquals);
-    if (logKeyIndex > -1) {
-      this.filter[logKey].splice(logKeyIndex, 1);
-    }
-  }
-
-  setFilterKeyword(keyword) {
-    this.filter.keyword = keyword;
+  setFilter({ logKey, valueEquals }) {
+    this.filter[logKey] = valueEquals;
```
Rob--W, line 19: `valueEquals` is longer an accurate description. Even something like this would be better:

```
setFilter(logKey, filterDescriptor) {
```

Maybe with a JSDoc comment before it to explain what the expected argument types are.

src/lib/web-component/filter-option/filter-option-element.js
```diff
@@ -27,32 +30,26 @@ class FilterOption extends HTMLElement {
...
-            detail: filterDetail,
-          })
-        );
+        this.dispatchFilterChangeEvent({ isNewFilterAdded: true });
       }
     }
   }
 
-  addNewCheckbox(labelText) {
+  addNewCheckbox(checkboxLabel) {
```
Rob--W, line 41: Both names were fine.
`labelText` wasn't wrong,
`checkboxLabel` isn't wrong either.

src/lib/web-component/filter-option/filter-option-element.js
```diff
@@ -27,32 +30,26 @@ class FilterOption extends HTMLElement {
...
     newCheckbox.querySelector('label span').textContent =
-      labelText || 'undefined';
+      checkboxLabel || 'undefined';
 
     const inputCheckbox = newCheckbox.querySelector('input');
-    inputCheckbox.value = labelText;
+    inputCheckbox.value = checkboxLabel;
     inputCheckbox.checked = true;
 
+    // adding to activeCheckboxLabels list since checkbox is checked
```
Rob--W, line 50: Nice. This is an example of a good comment. It explains why the code is like this, when it is not immediately obvious from the code.

src/lib/web-component/filter-option/filter-option-element.js
```diff
@@ -95,6 +82,29 @@ class FilterOption extends HTMLElement {
     this.toggleBtn.classList.toggle('expanded');
   }
 
+  handleCheckbox(checkboxLabel) {
```
Rob--W, line 85: I suggest to use `onCheckboxChange`. And also pass the value true or false, depending on whether the checkbox is checked, so you can use it instead of `.has(...)` in the if below.

src/lib/web-component/filter-option/filter-option-element.js
```diff
@@ -95,6 +82,29 @@ class FilterOption extends HTMLElement {
...
+      this.activeCheckboxLabels.add(checkboxLabel);
+    }
+    this.dispatchFilterChangeEvent({ isNewFilterAdded: false });
+  }
+
+  dispatchFilterChangeEvent({ isNewFilterAdded }) {
+    const filterDetail = {
+      filterObject: {
+        logKey: this.filterKey,
+        valueEquals: this.activeCheckboxLabels,
```
Rob--W, line 98: Make a copy of the `activeCheckboxLabels` Set so that the internal state cannot be changed by external listeners, and that external listeners won't unexpectedly have a mutated set when the internal state changes.

src/lib/ext-activitylog.js

tests/ext-activitylog.test.js
```diff
@@ -0,0 +1,52 @@
...
+  sendMessage.mockImplementation(() => {
+    return Promise.resolve({ existingLogs: [] });
+  });
+
+  const { activityLog } = new ActivityLog();
+  activityLog.handleNewLogs([log]);
+
+  expect(activityLog.isFilterMatched(log)).toBeTruthy();
+  expect(activityLog.isFilterMatched(logUnmatch)).toBeFalsy();
+});
```
Rob--W, line 52: I also suggest to have some unit tests to verify that the filter logic (of the model) works as intended. The logic may seem obvious now, but the implementation may change in the future, and then that person who refactors the implementation would greatly appreciate the unit tests that you have written.

tests/ext-activitylog.test.js
```diff
@@ -13,8 +14,24 @@ const domParser = new DOMParser();
...
+  </html>`,
+  { runScripts: 'dangerously' }
+);
+
 test('isFilterMatched function returns true when all properties of a log are matched with filter object, else it returns true', () => {
-  document.body.innerHTML = activityLogBody;
+  global.document.head.innerHTML = dom.window.document.head.innerHTML;
+  global.document.body.innerHTML = dom.window.document.body.innerHTML;
 
   const log = {
```
rpl, line 36: @atiqueahmedziad To make sure that the expected webcomponents are defined in the test document as expected we just need to import those modules.

E.g. we may export the custom elements classes in the related modules, something like:
```
export class FilterOption extends HTMLElement {
...
}
```
same for the LogView class.

Then in this test file we can just do something like the following (to load the modules and then explicitly check that the expected components are registered):
```
...
// Import the webcomponents modules 
import { FilterOption } from '../src/lib/web-component/filter-option/filter-option-element.js';
import { LogView } from '../src/lib/web-component/log-view/log-view-element.js';

...
const activityLogBody = domParser.parseFromString(activityLogHtml, 'text/html')
  .body.innerHTML;

test('isFilterMatched function returns true when all properties of a log are matched with filter object, else it returns true', () => {
  document.body.innerHTML = activityLogBody

  // Assert that the expected custom elements have been registered to the test window CustomElementRegistry.
  expect(window.customElements.get('log-view')).toBe(LogView);
  expect(window.customElements.get('filter-option')).toBe(FilterOption);

  // proceed with the rest of the test
  ...
```

src/lib/ext-activitylog.js
```diff
@@ -15,8 +15,14 @@ class Model {
     this.logs.push(...logs);
   }
 
-  setFilter({ logKey, valueEquals }) {
-    this.filter[logKey] = valueEquals;
+  /**
+   * Sets the values of any key in filter object
+   * @param {string} logKey - The key in the filter object.
+   * @param {Set} filterDescriptor - It contains the active checkboxes.
```
Rob--W, line 21: This description is too specific to the UI. The model doesn't need to know about how the filter is created. I intentionally chose a variable name (`filterDescriptor`) that clearly describes the role of the value, independent of the DOM.

The type definition is incorrect. For `logKey="keyword"`, `filterDescriptor` is expected to be a string. This mistake shows why the need for proper documentation is important: if even you, who wrote all code, can't get the type right at the first try, imagine the difficulty that others (who don't work on the code) will have at trying to figure out the parameters to pass.

The JSDoc here should be written such that anyone who isn't familiar with the whole code base is able to know what kind of values they are allowed to pass (and if not obvious - why). The main purpose is to help the reader with understanding how they can use the function.

src/lib/ext-activitylog.js
```diff
@@ -15,8 +15,14 @@ class Model {
...
+   */
+
```
Rob--W, line 23: Remove blank line between JSDoc comment and the function.

src/lib/ext-activitylog.js
```diff
@@ -177,7 +183,7 @@ class Controller {
 
   onFilterChange(filterDetail) {
     const { filterObject, isNewFilterAdded } = filterDetail;
-
+    //console.log(filterObject);
```
Rob--W, line 186: Remove debugging comment.

src/lib/web-component/filter-option/filter-option-element.js
```diff
@@ -85,6 +84,29 @@ class FilterOption extends HTMLElement {
...
+      this.activeCheckboxLabels.delete(checkboxLabel);
+    }
+    this.dispatchFilterChangeEvent({ isNewFilterAdded: false });
+  }
+
+  dispatchFilterChangeEvent({ isNewFilterAdded }) {
+    const filterDetail = {
+      filterObject: {
+        logKey: this.filterKey,
+        filterDescriptor: new Set([...this.activeCheckboxLabels]),
```
Rob--W, line 100: The `Set` constructor takes an iterable. A `Set` instance itself is also iterable, so you don't need to turn it into an array first.

`[...this.activeCheckboxLabels]` iterates over the set and results in an array.
`new Set([...this.activeCheckboxLabels])` results in two iterations: first to turn the set into an array, and then again the `Set` constructor iterates over the array to fill the set.

atiqueahmedziad: done :-)

src/lib/ext-activitylog.js
atiqueahmedziad: done.

src/lib/ext-activitylog.js
atiqueahmedziad: done.

src/lib/ext-activitylog.js
atiqueahmedziad: Thanks for explaining. I have updated the JSDoc description :-)

src/lib/ext-activitylog.js
```diff
@@ -4,25 +4,51 @@ class Model {
...
-    this.filter[logKey].push(valueEquals);
+  /**
+   * Sets the value of any key in filter
+   * @param {string} logKey - The key in the filter.
+   * @param {Set} filterDescriptor - It contains collection of string for any key except keyword key.
+   * @param {string} filterDescriptor - It contains string for keyword key.
+   */
+  setFilter({ logKey, filterDescriptor }) {
+    this.filter[logKey] = filterDescriptor;
   }
```
rpl, line 26: @atiqueahmedziad 
This jsdoc looks still wrong to me (e.g. there are two different definition of the filterDescriptor property, and if I would be a developer that is looking to this jsdoc to understand this function I would still be confused on how I should use it) and the current method signature does confuse me a bit (I had to follow the code to actually understand how that was going to be used).

In my opinion this method could take as its parameter an object with "the subset of the same properties of the filter object" that the caller want to update (with the exact same properties of the filter object as defined in the constructor al line 6, all properties would be optional to allow the caller to update a particular part of the filter and leave the rest unmodified).

I think that this would make the method easier to use and it should also make it easier to express it in the jsdoc in a non confusing way.

The method signature would be something like
```
setFilter(updatedFilterProps) {
  // here the code to merge updateFilterProps into this.filter
}
```

Let me know what do you think (and if my comment is clear enough)

atiqueahmedziad: Thanks @rpl for the idea. I have implemented this. :-)

src/lib/ext-activitylog.js
```diff
@@ -4,25 +4,50 @@ class Model {
...
   addNewLogs(logs) {
     this.logs.push(...logs);
   }
 
-  addFilter({ logKey, valueEquals }) {
-    this.filter[logKey].push(valueEquals);
+  /**
+   * Set the value of any key in filter
+   * @param {object} updateFilterProps - It contains key and value to update
+   * the filter.
```
rpl, line 21: @atiqueahmedziad I think that to make this jsdoc really useful you should actually spell what are the properties expected in that object, what is their type and what they represent, the current comment is quite generic.

As a "reader of the jsdoc" I would like to see explicitly that the expected properties are `id`, `viewType`, `type` and `keyword`, and then I would like to know what `id`, `viewType`, `type` and `keyword` represents and what is their expected types.

src/lib/web-component/filter-keyword/filter-keyword-element.js
```diff
@@ -0,0 +1,39 @@
...
+      .content.cloneNode(true);
+
+    this.inputBox = filterContainer.querySelector('input[name="keyword"]');
+
+    shadow.appendChild(filterContainer);
+  }
+
+  connectedCallback() {
+    let timer = 0;
+    this.inputBoxListener = (event) => {
```
Rob--W, line 17: The debouncing logic is quite repetitive (and has a bug - see my comment at the filter-option class).

Consider fixing the logic and putting the implementation at a shared location.

src/lib/web-component/filter-keyword/filter-keyword-element.js
```diff
@@ -0,0 +1,39 @@
+export class FilterKeyword extends HTMLElement {
+  constructor() {
+    super();
+
+    const shadow = this.attachShadow({ mode: 'open' });
+    const filterContainer = document
+      .querySelector('#filterKeywordTemplate')
+      .content.cloneNode(true);
+
```
Rob--W, line 9: `filter-keyword.css` is currently empty and not referenced. If you don't plan on using it, remove it from the PR.

src/lib/web-component/filter-option/filter-option-element.js
```diff
@@ -85,6 +84,28 @@ class FilterOption extends HTMLElement {
...
+    } else {
+      this.activeCheckboxLabels.delete(checkboxLabel);
+    }
+    this.dispatchFilterChangeEvent({ isNewFilterAdded: false });
+  }
+
+  dispatchFilterChangeEvent({ isNewFilterAdded }) {
+    const filterDetail = {
+      updateFilterProps: {
+        [this.filterKey]: new Set(this.activeCheckboxLabels),
```
Rob--W, line 99: Nice one, good use of modern JS syntax for readability!

src/lib/web-component/filter-keyword/filter-keyword-element.js
atiqueahmedziad: should I use the following delay function for this?
```
  delay(fn, ms) {
    let timer = 0;
    return function (...args) {
      clearTimeout(timer);
      timer = setTimeout(fn.bind(this, ...args), ms || 0);
    };
  }
```

Rob--W: https://github.com/mozilla/extension-activity-monitor/pull/14#discussion_r460060187 points to a bug. When disconnected, an element should cease to work. In the current implementation it will continue to trigger events.

A way to avoid this issue and to avoid duplication is to let the controller coalesce the repeated filter changes. It can update the filter, but debounce the filter change propagation until the value stabilizes. SInce this is an optimization that isn't required for this PR, I would not bother with it now, and defer it to later.

In the `keyword` input, there is likely some value in debouncing since a user types.
In the checkbox component, I would remove it, since there is rarely a use for debouncing checkbox changes.

So, to summarize: to keep this PR focused on what it should be doing, fix the debounce logic to clear the timer on element removal, and remove the debounce logic from the filter-option component.

src/lib/ext-activitylog.js
```diff
@@ -4,25 +4,58 @@ class Model {
...
 
-  addFilter({ logKey, valueEquals }) {
-    this.filter[logKey].push(valueEquals);
+  /**
+   * @param {object} updateFilter - It contains key and value to replace
+   * the filter
+   * @param {Set<string>} [updateFilter.id] - It contains the extension ids.
+   * @param {Set<string>} [updateFilter.viewType] - It contains view types that
+   * includes background, popup, sidebar, tab, devtools_page, devtools_panel.
+   * @param {undefined} [updateFilter.viewType] - when [updateFilter.type] is
```
Rob--W, line 24: Don't create separate definitions for the same property. You can use `Set<string|undefined>` to merge the types. If you feel a need to explain the `undefined` value, do so in the description.

src/lib/ext-activitylog.js
```diff
@@ -4,25 +4,58 @@ class Model {
...
+   * content_script.
+   * @param {Set<string>} [updateFilter.type] - It contains api types that
+   * includes api_call, api_event, content_script, user_script.
+   * @param {string} [updateFilter.keyword] - It contains a string for
```
Rob--W, line 28: The description doesn't say more than the type. If you can't think of anything meaningful, just leave the description empty.

src/lib/web-component/filter-keyword/filter-keyword-element.js
```diff
@@ -0,0 +1,39 @@
...
+    this.inputBoxListener = (event) => {
+      clearTimeout(this.timer);
+      this.timer = setTimeout(() => {
+        const filterDetail = {
+          updateFilter: { keyword: event.target.inputBox.value },
+        };
+
+        this.dispatchEvent(
+          new CustomEvent('filterchange', { detail: filterDetail })
+        );
```
Rob--W, line 22: Move the logic to a `dispatchFilterChanged` method. You don't need to use `event.target.inputBox`, `this.inputBox` suffices.

src/lib/ext-activitylog.js

PR #15, opened 2020-07-21
feat: styles of log detail view is updated

atiqueahmedziad:
As suggested, in last meeting, I am adding `<pre>` in viewing log details in sidebar.

### DIFF
--- a/src/activitylog/activitylog.html
+++ b/src/activitylog/activitylog.html
@@ -53,7 +53,7 @@
           </table>
         </div>
         <div class="log-detail-wrapper" hidden>
-          <pre class="log-details"></pre>
+          <div class="log-details"></div>
           <div class="close"></div>
         </div>
       </div>
--- a/src/activitylog/activitylog.html
+++ b/src/activitylog/activitylog.html
@@ -53,7 +53,7 @@
           </table>
         </div>
         <div class="log-detail-wrapper" hidden>
-          <div class="log-details"></div>
+          <pre class="log-details"></pre>
           <div class="close"></div>
         </div>
       </div>
--- a/src/lib/web-component/log-view/log-view-element.js
+++ b/src/lib/web-component/log-view/log-view-element.js
@@ -52,7 +52,7 @@ class LogView extends HTMLElement {
   }
 
   openDetailSidebar(logDetails) {
-    const logString = JSON.stringify(logDetails);
+    const logString = JSON.stringify(logDetails, null, 4);
     this.logDetails.textContent = logString;
     this.logTableWrapper.classList.add('width-60');
     this.logDetailWrapper.hidden = false;
--- a/src/lib/web-component/log-view/log-view-element.js
+++ b/src/lib/web-component/log-view/log-view-element.js
@@ -52,7 +52,7 @@ class LogView extends HTMLElement {
   }
 
   openDetailSidebar(logDetails) {
-    const logString = JSON.stringify(logDetails, null, 4);
+    const logString = JSON.stringify(logDetails, null, 2);
     this.logDetails.textContent = logString;
     this.logTableWrapper.classList.add('width-60');
     this.logDetailWrapper.hidden = false;
--- a/src/lib/web-component/log-view/log-view.css
+++ b/src/lib/web-component/log-view/log-view.css
@@ -20,6 +20,7 @@
 
 .log-details {
   margin: 50px 15px 10px;
+  white-space: pre-wrap;
 }
 
 .width-60 {
--- a/src/lib/web-component/log-view/log-view.css
+++ b/src/lib/web-component/log-view/log-view.css
@@ -8,7 +8,7 @@
 }
 
 .log-table-wrapper {
-  overflow-y: scroll;
+  overflow-y: auto;
   max-height: 500px;
   height: auto;
 }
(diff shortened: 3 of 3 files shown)


src/lib/web-component/log-view/log-view.css
```diff
@@ -20,6 +20,7 @@
 
 .log-details {
   margin: 50px 15px 10px;
+  white-space: pre-wrap;
```
rpl, line 23: Given that you are using the CSS to make the element to be treat the whitespaces as a pre element, then you can leave the `.log-details` element as a `div` (in the end this is just a temporary workaround to make the log details a bit more readable in the short term, to be just used until you will be able to spent some time on making the sidebar details more complete and useful).

To make the log details a bit more readable, along with this temporary styling change you should also change the related `JSON.stringify` call to make it "pretty print" (indent it basically) a bit the JSON data in the generated string (If you have never used the other `JSON.stringify` parameters before you can take at this doc page: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/stringify).

I guess that there may be some other css tweaks to make the content fit well into the page layout, e.g. you could probably make the wrapper element to show a scrollbar if the content is overflowing the wrapper element, but don't spend too much more time than that in this PR, this is just a temporary measure (and the activity log page in general could likely use some more work on its CSS to improve its layout, but it can definitely be deferred to a separate PR).

src/activitylog/activitylog.html
```diff
@@ -53,7 +53,7 @@ <h2>Activity Logs</h2>
           </table>
         </div>
         <div class="log-detail-wrapper" hidden>
-          <div class="log-details"></div>
+          <pre class="log-details"></pre>
```
rpl, line 56: As mentioned in the other comment, let's change this back to a div, the CSS styling should be already enough.

src/lib/web-component/log-view/log-view.css
atiqueahmedziad: done. Thanks :-).

src/lib/web-component/log-view/log-view-element.js
```diff
@@ -52,7 +52,7 @@ class LogView extends HTMLElement {
   }
 
   openDetailSidebar(logDetails) {
-    const logString = JSON.stringify(logDetails);
+    const logString = JSON.stringify(logDetails, null, 4);
```
rpl, line 55: Nit, 2 (instead of 4) may be enough for the indentation, and it would spare a bit of horizontal space.

src/lib/web-component/log-view/log-view.css
```diff
@@ -19,7 +24,11 @@
 }
 
 .log-details {
-  margin: 50px 15px 10px;
+  margin: 40px 0 10px;
+  white-space: pre-wrap;
+  max-height: 450px;
+  overflow-y: scroll;
```
rpl, line 30: have you checked what happens if the json log contains something (e.g. a very long string value or a very long object property name)? would it has to also scroll horizontally? would setting the overflow to `auto` also work and made the scrollbar hidden when the json fits into the available space?

src/lib/web-component/log-view/log-view.css
```diff
@@ -1,5 +1,10 @@
+* {
+  box-sizing: border-box;
+}
+
 .activity-log-container {
   position: relative;
+  margin-bottom: 30px;
```
atiqueahmedziad, line 7: This would add some margin under the table.

src/lib/web-component/log-view/log-view.css
```diff
@@ -19,7 +24,11 @@
 }
 
 .log-details {
-  margin: 50px 15px 10px;
+  margin: 40px 0 10px;
+  white-space: pre-wrap;
+  max-height: 450px;
+  overflow-y: scroll;
+  padding-left: 20px;
```
atiqueahmedziad, line 31: Indenting the layout a bit from left so that log string doesn't align too closely with the edge.

src/lib/web-component/log-view/log-view.css
atiqueahmedziad: When  there us a log string value, it gets indent into the next time. Have a look at the screenshot.
### screenshot
Screenshot 2020-07-28 at 11 49 08 PM

src/lib/web-component/log-view/log-view.css
```diff
@@ -19,7 +24,11 @@
 }
 
 .log-details {
-  margin: 50px 15px 10px;
+  margin: 40px 0 10px;
+  white-space: pre-wrap;
+  max-height: 450px;
```
atiqueahmedziad, line 29: Setting the `max-height` to 450px since I used 50px pixels margin in total at [line 27](https://github.com/mozilla/extension-activity-monitor/pull/15/files#diff-3e34fcdde14c53a51242dcc1edf18813R27) (40px top and 10px bottom) to give it box like look.

src/lib/web-component/log-view/log-view.css
atiqueahmedziad: > would setting the overflow to `auto` also work and made the scrollbar hidden when the json fits into the available space?

Yeah, I have updated the overflow to be `auto` :-)

PR #28, opened 2020-08-02
feat: load logs from json file

### LINKED ISSUE #25: loading and showing the logs from a file
The load logs funationality has been added on pr #2. This issue is for -
- create an input field to load a log file from the PC.
- renders the logs from the given file into a table view.

atiqueahmedziad:
Fixes #25 
### The following is done:
- Load log file (json) from local machine
- Loaded logs will be shown in `log-view` opening a new tab with the following parameter at the end of url: `?page=loadLog&file=filename.json`.
- The loaded logs page has a page title set to `Loaded Logs - fileName.json`. This will help to identify it's a loaded logs page.
- There is a button named "Clear All Loaded Logs" which clears all the loaded logs. User can access their loaded logs at any time with respective parameters in URL until they clear the loaded logs.
- [`menu-container`](https://github.com/mozilla/extension-activity-monitor/pull/28/files#diff-f63ec68852d55aaede19b261da7a08d6R25) is hidden in `load log` page. That means, there will be no option to save, clear, load logs and clear loaded logs in a loaded logs page.

### Screenshot
Screenshot 2020-08-02 at 10 14 24 PM

### DIFF
--- a/src/lib/ext-activitylog.js
+++ b/src/lib/ext-activitylog.js
@@ -158,27 +158,27 @@ class Controller {
     this.view.extFilter.addEventListener('filterchange', this);
     this.view.viewTypeFilter.addEventListener('filterchange', this);
     this.view.apiTypeFilter.addEventListener('filterchange', this);
-    const searchParams = window.location.search;
 
-    if (searchParams.includes('page=loadLog&file=')) {
+    const searchParams = new URLSearchParams(
+      document.location.search.substring(1)
+    );
+    const fileName = searchParams.get('file');
+
+    if (fileName) {
       this.view.menuContainer.hidden = true;
-
-      const fileName = searchParams.substring(
-        searchParams.indexOf('file=') + 5
-      );
-      if (!fileName) {
-        return;
-      }
-
       document.title = `Loaded Logs - ${fileName}`;
 
-      const logs = await browser.runtime.sendMessage({
-        requestType: 'getLoadedLogs',
-        requestTo: 'ext-monitor',
-        detail: { fileName },
-      });
+      const logs = await browser.runtime
+        .sendMessage({
+          requestType: 'getLoadedLogs',
+          requestTo: 'ext-monitor',
+          detail: { fileName },
+        })
+        .catch((error) => {
+          this.view.setError(error.message);
+        });
 
-      if (logs.length) {
+      if (logs?.length) {
         this.handleNewLogs(logs);
       }
     } else {
@@ -251,10 +251,10 @@ class Controller {
 
       const storedFileName = await browser.runtime.sendMessage({
         requestTo: 'ext-monitor',
-        requestType: 'loadLogs',
-        detail: { logs, fileName: file.name },
+        requestType: 'setLoadedLogs',
+        detail: { fileName: file.name, logs },
       });
-      const searchParams = `page=loadLog&file=${storedFileName}`;
+      const searchParams = `file=${storedFileName}`;
 
       this.view.setError(null);
 
--- a/src/lib/ext-activitylog.js
+++ b/src/lib/ext-activitylog.js
@@ -1,4 +1,5 @@
-import { save } from './save-load.js';
+import { getActivityLogPageURL } from './ext-listen.js';
+import { load, save } from './save-load.js';
 
 class Model {
   constructor() {
@@ -64,8 +65,11 @@ class Model {
 class View {
   constructor() {
     this.logView = document.querySelector('log-view');
+    this.menuContainer = document.querySelector('.menu-container');
     this.clearLogBtn = document.querySelector('#clearLogBtn');
     this.saveLogBtn = document.querySelector('#saveLogBtn');
+    this.loadLogFile = document.querySelector('input[name="loadLogFile"]');
+    this.clearLoadedLogBtn = document.querySelector('#clearLoadedLogs');
     this.notice = document.querySelector('.notice');
 
     this.extFilter = document.querySelector('filter-option[filter-key="id"]');
@@ -79,6 +83,8 @@ class View {
 
     this.clearLogBtn.addEventListener('click', this);
     this.saveLogBtn.addEventListener('click', this);
+    this.loadLogFile.addEventListener('change', this);
+    this.clearLoadedLogBtn.addEventListener('click', this);
   }
 
   setLogFilter(filterFunc) {
@@ -94,9 +100,20 @@ class View {
         case this.clearLogBtn:
           this.clearLogBtn.dispatchEvent(new CustomEvent('clearlog'));
           break;
+        case this.clearLoadedLogBtn:
+          this.clearLoadedLogBtn.dispatchEvent(
+            new CustomEvent('clearloadedlog')
+          );
+          break;
         default:
           throw new Error(`wrong event target - ${event.target.tagName}`);
       }
+    } else if (event.type === 'change' && event.target === this.loadLogFile) {
+      const logFile = event.target.files[0];
+      this.loadLogFile.value = '';
+      this.loadLogFile.dispatchEvent(
+        new CustomEvent('loadlog', { detail: logFile })
+      );
     } else {
       throw new Error(`wrong event type - ${event.type}`);
     }
@@ -137,31 +154,58 @@ class Controller {
   }
 
   async init() {
-    this.view.clearLogBtn.addEventListener('clearlog', this);
-    this.view.saveLogBtn.addEventListener('savelog', this);
     this.view.keywordFilter.addEventListener('filterchange', this);
     this.view.extFilter.addEventListener('filterchange', this);
     this.view.viewTypeFilter.addEventListener('filterchange', this);
     this.view.apiTypeFilter.addEventListener('filterchange', this);
+    const searchParams = window.location.search;
 
-    browser.runtime.onMessage.addListener((message) => {
-      const { requestTo, requestType } = message;
+    if (searchParams.includes('page=loadLog&file=')) {
+      this.view.menuContainer.hidden = true;
 
-      if (requestTo !== 'activity-log') {
+      const fileName = searchParams.substring(
+        searchParams.indexOf('file=') + 5
+      );
+      if (!fileName) {
         return;
       }
 
-      if (requestType === 'appendLogs') {
-        this.handleNewLogs([message.log]);
-      } else {
-        throw new Error(`wrong request type found - ${requestType}`);
+      document.title = `Loaded Logs - ${fileName}`;
+
+      const logs = await browser.runtime.sendMessage({
+        requestType: 'getLoadedLogs',
+        requestTo: 'ext-monitor',
+        detail: { fileName },
+      });
+
+      if (logs.length) {
+        this.handleNewLogs(logs);
       }
-    });
+    } else {
+      this.view.loadLogFile.addEventListener('loadlog', this);
+      this.view.clearLogBtn.addEventListener('clearlog', this);
+      this.view.saveLogBtn.addEventListener('savelog', this);
+      this.view.clearLoadedLogBtn.addEventListener('clearloadedlog', this);
 
-    const existingLogs = await this.getExistingLogs();
+      browser.runtime.onMessage.addListener((message) => {
+        const { requestTo, requestType } = message;
 
-    if (existingLogs.length) {
-      this.handleNewLogs(existingLogs);
+        if (requestTo !== 'activity-log') {
+          return;
+        }
+
+        if (requestType === 'appendLogs') {
+          this.handleNewLogs([message.log]);
+        } else {
+          throw new Error(`wrong request type found - ${requestType}`);
+        }
+      });
+
+      const existingLogs = await this.getExistingLogs();
+
+      if (existingLogs.length) {
+        this.handleNewLogs(existingLogs);
+      }
     }
   }
 
@@ -180,6 +224,9 @@ class Controller {
 
   handleEvent(event) {
     switch (event.type) {
+      case 'loadlog':
+        this.loadLogs(event.detail);
+        break;
       case 'savelog':
         this.saveLogs();
         break;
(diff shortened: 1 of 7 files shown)


src/lib/ext-activitylog.js
```diff
@@ -137,31 +154,58 @@ class Controller {
...
+    if (searchParams.includes('page=loadLog&file=')) {
+      this.view.menuContainer.hidden = true;
 
-      if (requestTo !== 'activity-log') {
+      const fileName = searchParams.substring(
+        searchParams.indexOf('file=') + 5
+      );
+      if (!fileName) {
         return;
       }
```
rpl, line 171: We could use URLSearchParams to more easily parse the URL search params, see https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams/get.

Also, it may be better to make the condition we check in the `if` at line 163 to do not depend from the order of the search params (I mean that we may not need the `page=loadLog` search param and the the `searchParams.includes(...)` condition could be replaced by checking that we do have a `file` search param).

src/lib/ext-activitylog.js
```diff
@@ -64,8 +65,11 @@ class Model {
 class View {
   constructor() {
     this.logView = document.querySelector('log-view');
+    this.menuContainer = document.querySelector('.menu-container');
     this.clearLogBtn = document.querySelector('#clearLogBtn');
     this.saveLogBtn = document.querySelector('#saveLogBtn');
+    this.loadLogFile = document.querySelector('input[name="loadLogFile"]');
+    this.clearLoadedLogBtn = document.querySelector('#clearLoadedLogs');
```
rpl, line 72: If the activitylog.html page for a loaded log file is only being used to review the logs in the loaded file, we may not need to different clear logs buttons, I have the feeling that it may just be confusing and the user may clear the logs being collected by mistake if it clicks the wrong button.

src/lib/ext-monitor.js
```diff
@@ -1,6 +1,10 @@
 import { getActivityLogPageURL } from './ext-listen.js';
 
 export default class ExtensionMonitor {
+  /**
+   * @type {Array.<{logs: Array<Object>, fileName: String}>}
+   */
+  loadedLogs = [];
```
rpl, line 7: Similar to my other comment, if the activitylog.html page opened for a loaded log file is only being used to review those logs, we may not need both `loadedLogs` and `logs` (eg. that instance of the activitylog.html page is not meant to subscribe to receive realtime logs at all).

atiqueahmedziad: @rpl here the `loadedLogs` array is in `ext-monitor.js` (background) where I am storing the loaded logs along with the filename. And when I am opening a new tab with parameter like: `?page=loadLog&file=fileName.json`, the activitylog.html getting the loaded logs from the background and [renders the logs in the table view](https://github.com/mozilla/extension-activity-monitor/pull/28/files#diff-c00fd244576d6e063a30656f921d4db6R182)

I am storing the loaded logs in the background since, if a user mistakenly closes the tab / reloads the tab, he doesn't lose the loaded logs. If he mistakenly closes the tab, he will still be able to see the loaded logs adding the URL parameters: `?page=loadLog&file=fileName.json`.

rpl: @atiqueahmedziad ah, I see. Thanks for pointing it out!

Wouldn't be reasonable to store the logs in a map keyed by the filename string?

atiqueahmedziad: Yeah, it's reasonsable :-) Thanks.
I have updated the PR storing the logs and file name in a Map.

src/lib/ext-activitylog.js
atiqueahmedziad: Thanks, it's done :-)

src/lib/ext-activitylog.js
atiqueahmedziad: @rpl Should we clear the loaded logs when a user clicks on clear logs buttons?

rpl: If I recall correctly, when we discussed about this during the last meeting we agreed that clearing the loaded logs may not make sense.

Nevertheless, given that we are storing those logs in the background page, at some point we should be removing them, otherwise we may keep all of the loaded files around until Firefox is being restarted.

As a user, I would expect to "close the loaded file" more than "clear the loaded logs", but from an implementation perspective they are basically the same thing, just named differently to match what is more likely the user expectation.

And so (I'm basically thinking about this aloud) I feel that, when the activitylog page is this kind of "view logs from file" mode, it may be reasonable to:

- list the logs loaded from file (maybe in a select input element), and allow the user to switch between the loaded logs files
- have a button to "close the loaded logs file", which would remove the entry for that loaded logs from the background page

@wagnerand what do you think about this? would the behavior described above match your expectation as a user
(also @Rob--W for his opinion on this).

wagnerand: I think expecting a user to have to click on _Close the loaded logs file_ might be a bit too much. It's not unlikely they forget to click the button and just close the tab, in which case the logs would not be removed from memory. I could think about 2 possible options to do this:

Option 1 (simple): When the tab gets closed, remove the logs from the background page. If the user decides to undo closing the tab, they have to reload the file.

Option 2 (a bit more involved): When the tab gets closed, start a timer (timeout tbd, but I guess 1 minute or so would do). When it times out, check if the log is loaded in a tab. If so, just do nothing. If not, remove the log from the background page.

Of course, the check whether a log is still shown somewhere, could be something else, for example a lock we set on the log data when the file gets loaded, and unset when the tab gets closed. After the timeout, we'd check whether the lock is set and remove the log from background page if it's not.

src/lib/ext-activitylog.js
```diff
@@ -194,6 +233,34 @@ class Controller {
     }
   }
 
+  async loadLogs(file) {
+    try {
+      const logStr = await load.loadLogAsText(file);
+      const logs = JSON.parse(logStr);
```
rpl, line 239: Nit, we do have a load ESM module, I think that JSON.parse should be a responsibility of that module instead of being a responsibility of the Controller class.

src/lib/ext-activitylog.js
```diff
@@ -194,6 +233,34 @@ class Controller {
...
+
+      const tab = await browser.tabs.create({
+        url: getActivityLogPageURL(searchParams),
+      });
+
+      await browser.runtime.sendMessage({
+        requestTo: 'ext-monitor',
+        requestType: 'setLoadedLogsTabId',
+        detail: { tabId: tab.id, fileName: storedFileName },
+      });
```
rpl, line 258: Instead on being the activitylog page to send the logs to the background page, then create a new tab and then send another message to the background page to associate the tab.id to the loaded file, this could actually be a responsibility of the background page (or the ext-monitor module actually), as part of what it does to handle a single 'openLoadedLogsTab`.

(Let me know if my request isn't clear enough as described above)

src/lib/ext-monitor.js
```diff
@@ -1,6 +1,10 @@
 import { getActivityLogPageURL } from './ext-listen.js';
 
 export default class ExtensionMonitor {
+  // Map<string, Array>
+  loadedLogs = new Map();
+  // Map<number, string>
+  loadedLogsTabIds = new Map();
```
rpl, line 7: Given that we agreed to clear the logs automatically when the tab we have opened has been navigated or closed, it may be better to:

- keep the loaded logs in a map keyed by tabId, as the value of the map we could store an object that contains both the filename and the loaded logs in two separate properties

does that sounds reasonable?

(this should, at least at a first glance, also simplify a bit `setLoadedLogs`, e.g. we will not need to care about the unique file name thing)

src/lib/ext-monitor.js
```diff
@@ -76,17 +80,49 @@ export default class ExtensionMonitor {
...
+    return logs;
+  }
+
+  setLoadedLogsTabId({ fileName, tabId }) {
+    this.loadedLogsTabIds.set(tabId, fileName);
+  }
+
   messageListener = (message) => {
-    const { requestType, requestTo } = message;
+    const { requestType, requestTo, detail } = message;
```
rpl, line 118: Let's rename detail to something like `requestParams`, in the end it is what we pass to the requestType method and so it should be easier to understand what that property is.

src/lib/ext-activitylog.js
atiqueahmedziad: So, in summary, I will be doing the following - 
- Send the loaded file from activitylog.html to the background.
- In the background, I will -
    - read the log file. 
    - open a new tab.
    - load the logs in that tab.

@rpl Am I right?

src/lib/ext-monitor.js
atiqueahmedziad: Yeah, sounds good :-)

src/lib/ext-monitor.js
atiqueahmedziad: Thanks for the suggestion. Updated the name :-)

src/lib/ext-activitylog.js

tests/ext-monitor.test.js
```diff
@@ -234,17 +234,19 @@ describe('messageListeners functionalities test', () => {
...
     runtime: { onMessage: { addListener } },
+    tabs: { onRemoved: { addListener } },
   };
 
   const extMonitor = new ExtensionMonitor();
   extMonitor.init();
 
-  expect(addListener).toHaveBeenCalled();
+  // runtime.onMessage.addListener and tabs.onRemoved.addListener are called.
+  expect(addListener).toHaveBeenCalledTimes(2);
```
rpl, line 249: the change applied to the test is workarounding the failure instead of aligning the test to the expected new behavior, as an example the test would not fail if the ext-monitor module does subscribe two runtime.onMessage listeners or two tabs.onRemoved listeners but not both as expected.

you should create two separate mocked addListener functions and then assert that each one of them has been called exactly once.

src/lib/ext-monitor.js
```diff
@@ -97,7 +123,15 @@ export default class ExtensionMonitor {
     }
   };
 
+  onRemovedListener = (tabId) => {
+    // When we close any tab which was loaded with logs from a log file.
```
rpl, line 127: Nit, `// Remove the loaded logs related to the closed tab` may be slightly more clear and as short as yours.

src/lib/ext-monitor.js
```diff
@@ -1,6 +1,9 @@
 import { getActivityLogPageURL } from './ext-listen.js';
+import { load } from './save-load.js';
 
 export default class ExtensionMonitor {
+  // Map<number, Array>
+  loadedLogs = new Map();
```
rpl, line 6: The inline comments does explicitly mention what is the type of the key, but not what it does represents, and the name of the property does not make it clear neither.

Calling this property `loadedLogsByTabId` would be making it clear enough without having to spelling that again in the inline comment.
The property is used only in 4 places and so even if the name is a bit longer it shouldn't make the core too much more verbose.

Otherwise you should at least mention it nearby the inline comment.

src/lib/ext-activitylog.js
rpl: yeah that's right (but I notice that you did already proceed to implement it, and so my reply did come late :-P)

src/lib/ext-monitor.js
atiqueahmedziad: done :-)

tests/ext-monitor.test.js
```diff
@@ -243,10 +243,13 @@ test('listeners are registered at initialization', () => {
...
   const extMonitor = new ExtensionMonitor();
+  const messageListenerFn = jest.spyOn(extMonitor, 'messageListener');
+  const onRemovedListenerFn = jest.spyOn(extMonitor, 'onRemovedListener');
+
   extMonitor.init();
 
-  // runtime.onMessage.addListener and tabs.onRemoved.addListener are called.
-  expect(addListener).toHaveBeenCalledTimes(2);
+  expect(addListener).toHaveBeenNthCalledWith(1, messageListenerFn);
+  expect(addListener).toHaveBeenNthCalledWith(2, onRemovedListenerFn);
```
atiqueahmedziad, line 252: @rpl how is this approach to you ?

rpl: > @rpl how is this approach to you ?
> **update**:
> I have checked like the following:
> 
> ```
> expect(addListener).toHaveBeenCalledTimes(2);
> expect(addListener).toHaveBeenNthCalledWith(1, messageListenerFn);
> expect(addListener).toHaveBeenNthCalledWith(2, onRemovedListenerFn);
> ```

Even with that change, the test does still pass if ext-monitor's `init` method does call (by mistake) twice `browser.tabs.onRemoved.addListener` on both the functions instead of calling `browser.runtime.onMessage.addListener` with `messageListenerFn` and `browser.tabs.onRemoved.addListener` with `onRemovedListenerFn`.

In practice, if ext-monitor's `init` does look like the following then the test case is supposed to fail:
```
  init() {
    browser.tabs.onRemoved.addListener(this.messageListener);
    browser.tabs.onRemoved.addListener(this.onRemovedListener);
  }
```

The current version doesn't.

If you just use two separate mocks for the two `addListener` methods, the test would fail as expected:
```
  const runtimeMessageAddListener = jest.fn();
  const tabsRemovedAddListener = jest.fn();

  window.browser = {
    runtime: { onMessage: { addListener: runtimeMessageAddListener } },
    tabs: { onRemoved: { addListener: tabsRemovedAddListener } },
  };

....
```

atiqueahmedziad:
Thanks for the explanation @rpl . I have updated the test :-)

rpl:
@atiqueahmedziad this PR looks good, while trying it locally I did notice another details, but we can deal with it in a follow up (to discuss about it and agree on our preference):

There is a difference between how the timestamp are formatted in the log view while "collected in realtime" vs. "when they are being read from a loaded JSON file":

diff_between_timestamps_in_realtime_vs_loadedlogs

would you mind to file an issue about this?

atiqueahmedziad:
> would you mind to file an issue about this?

Yeah, I am going to file an issue. Thanks for noticing.

Issue #27, opened 2020-08-02
Save logs with/without filter applied 


atiqueahmedziad:
Current when saving logs, it saves all the logs that have been captured despite any filter had been applied or not. It should show ask the user whether to save all the logs or only filtered logs (if filters are applied).

Rob--W:
Related to #24, nice-to-have.

PR #31, opened 2020-08-07
feat: ActivityLog DevTools panel added and Filter by tabId is supported

### LINKED ISSUE #22: ActivityLog DevTools panel
As being discussed in out last meeting, to make it easier to look to logs related to a particular tab we could register a devtools panel and load the activity log page into it.

When the activitylog page is loaded as a devtools panel the logs should be filtered automatically by tab id based on the value of the `browser.devtools.inspectedWindow.tabId`.

atiqueahmedziad:
Fixes #22 

- It creates a new panel named "Extension Activity" in the developer toolbox. 
- tabId is retrieved when the developer toolbox has been opened and generated the activity log page url link `activitylog.html?filterTabId=tabId`
- the activitylog page is loaded lazily when the user selects the panel.

### DIFF
--- a/src/lib/ext-activitylog.js
+++ b/src/lib/ext-activitylog.js
@@ -198,8 +198,8 @@ class View {
     this.logView.clearTable();
   }
 
-  updateLogHeading(newHeading) {
-    this.logHeading.textContent = newHeading;
+  renderView({ filterTabId }) {
+    this.logHeading.textContent = `Activity Logs Filtered By Tab Id: ${filterTabId}`;
   }
 }
 
@@ -272,9 +272,7 @@ class Controller {
       }
 
       if (filterTabId) {
-        this.view.updateLogHeading(
-          `Activity Logs Filtered By Tab Id: ${filterTabId}`
-        );
+        this.view.renderView({ filterTabId });
 
         const filterDetail = {
           updateFilter: { tabId: filterTabId },
--- a/src/lib/ext-activitylog.js
+++ b/src/lib/ext-activitylog.js
@@ -100,7 +100,7 @@ class Model {
   }
 
   matchFilterTabId({ tabId }) {
-    if (this.filter.tabId === null) {
+    if (this.filter.tabId == null) {
       return true;
     }
 
@@ -197,6 +197,10 @@ class View {
   clearTable() {
     this.logView.clearTable();
   }
+
+  updateLogHeading(newHeading) {
+    this.logHeading.textContent = newHeading;
+  }
 }
 
 class Controller {
@@ -268,7 +272,9 @@ class Controller {
       }
 
       if (filterTabId) {
-        this.view.logHeading.textContent = `${this.view.logHeading.textContent} Filtered By Tab Id: ${filterTabId}`;
+        this.view.updateLogHeading(
+          `Activity Logs Filtered By Tab Id: ${filterTabId}`
+        );
 
         const filterDetail = {
           updateFilter: { tabId: filterTabId },
@@ -325,7 +331,7 @@ class Controller {
 
   async saveLogs() {
     try {
-      await save.saveAsJSON(this.model.logs);
+      await save.saveAsJSON();
       this.view.setError(null);
     } catch (error) {
       this.view.setError(error.message);
--- a/src/lib/ext-activitylog.js
+++ b/src/lib/ext-activitylog.js
@@ -8,6 +8,7 @@ class Model {
       viewType: new Set(),
       type: new Set(),
       name: new Set(),
+      tabId: null,
       keyword: '',
       timeStamp: null,
     };
@@ -32,6 +33,8 @@ class Model {
    * filter is not applied.
    * @param {number} [updateFilter.timeStamp.start]
    * @param {number} [updateFilter.timeStamp.stop]
+   * @param {null|number} [updateFilter.tabId] - It's null when search paramter
+   * `filterTabId` is not used.
    */
   setFilter(updateFilter) {
     Object.assign(this.filter, updateFilter);
@@ -44,7 +47,8 @@ class Model {
       this.matchFilterType(log.type) &&
       this.matchFilterApiName(log) &&
       this.matchFilterKeyword(log.data) &&
-      this.matchFilterTimestamp(log.timeStamp)
+      this.matchFilterTimestamp(log.timeStamp) &&
+      this.matchFilterTabId(log.data)
     );
   }
 
@@ -93,6 +97,18 @@ class Model {
     }
 
     return true;
+  }
+
+  matchFilterTabId({ tabId }) {
+    if (!this.filter.tabId) {
+      return true;
+    }
+
+    if (tabId && tabId === this.filter.tabId) {
+      return true;
+    }
+
+    return false;
   }
 
   clearLogs() {
@@ -206,6 +222,7 @@ class Controller {
       document.location.search.substring(1)
     );
     const fileName = searchParams.get('file');
+    const filterTabId = parseInt(searchParams.get('filterTabId'), 10);
 
     if (fileName) {
       this.view.menuContainer.hidden = true;
@@ -228,10 +245,6 @@ class Controller {
         this.handleNewLogs(logs);
       }
     } else {
-      this.view.loadLogFile.addEventListener('loadlog', this);
-      this.view.clearLogBtn.addEventListener('clearlog', this);
-      this.view.saveLogBtn.addEventListener('savelog', this);
-
       browser.runtime.onMessage.addListener((message) => {
         const { requestTo, requestType } = message;
 
@@ -251,6 +264,23 @@ class Controller {
       if (existingLogs.length) {
         this.handleNewLogs(existingLogs);
       }
+
+      if (filterTabId) {
+        // currently menu container is hidden when activity log page is
+        // filtered by tab id
+        this.view.menuContainer.hidden = true;
+
+        const filterDetail = {
+          updateFilter: { tabId: filterTabId },
+        };
+
+        this.onFilterChange(filterDetail);
+        return;
+      }
+
+      this.view.loadLogFile.addEventListener('loadlog', this);
+      this.view.clearLogBtn.addEventListener('clearlog', this);
+      this.view.saveLogBtn.addEventListener('savelog', this);
     }
   }
 
--- a/src/lib/ext-activitylog.js
+++ b/src/lib/ext-activitylog.js
@@ -100,15 +100,11 @@ class Model {
   }
 
   matchFilterTabId({ tabId }) {
-    if (!this.filter.tabId) {
+    if (this.filter.tabId === null) {
       return true;
     }
 
-    if (tabId && tabId === this.filter.tabId) {
-      return true;
-    }
-
-    return false;
+    return this.filter.tabId === tabId;
   }
 
   clearLogs() {
--- a/src/lib/ext-activitylog.js
+++ b/src/lib/ext-activitylog.js
@@ -242,6 +242,7 @@ class Controller {
         this.handleNewLogs(logs);
       }
     } else {
+      browser.runtime.connect({ name: 'monitor-realtime-logs' });
       browser.runtime.onMessage.addListener((message) => {
         const { requestTo, requestType } = message;
 
--- a/src/lib/ext-activitylog.js
+++ b/src/lib/ext-activitylog.js
@@ -242,6 +242,10 @@ class Controller {
         this.handleNewLogs(logs);
       }
     } else {
+      this.view.loadLogFile.addEventListener('loadlog', this);
+      this.view.clearLogBtn.addEventListener('clearlog', this);
+      this.view.saveLogBtn.addEventListener('savelog', this);
+
       browser.runtime.connect({ name: 'monitor-realtime-logs' });
       browser.runtime.onMessage.addListener((message) => {
         const { requestTo, requestType } = message;
@@ -272,10 +276,6 @@ class Controller {
 
         this.onFilterChange(filterDetail);
       }
-
-      this.view.loadLogFile.addEventListener('loadlog', this);
-      this.view.clearLogBtn.addEventListener('clearlog', this);
-      this.view.saveLogBtn.addEventListener('savelog', this);
     }
   }
 
--- a/src/lib/ext-activitylog.js
+++ b/src/lib/ext-activitylog.js
@@ -198,7 +198,7 @@ class View {
     this.logView.clearTable();
   }
 
-  renderView({ filterTabId }) {
+  renderHeading({ filterTabId }) {
     this.logHeading.textContent = `Activity Logs Filtered By Tab Id: ${filterTabId}`;
   }
 }
@@ -272,7 +272,7 @@ class Controller {
       }
 
       if (filterTabId) {
-        this.view.renderView({ filterTabId });
+        this.view.renderHeading({ filterTabId });
 
         const filterDetail = {
           updateFilter: { tabId: filterTabId },
--- a/src/lib/ext-activitylog.js
+++ b/src/lib/ext-activitylog.js
@@ -270,7 +270,6 @@ class Controller {
         };
 
         this.onFilterChange(filterDetail);
-        return;
       }
 
       this.view.loadLogFile.addEventListener('loadlog', this);
--- a/src/lib/ext-activitylog.js
+++ b/src/lib/ext-activitylog.js
@@ -120,6 +120,7 @@ class View {
     this.saveLogBtn = document.querySelector('#saveLogBtn');
     this.loadLogFile = document.querySelector('input[name="loadLogFile"]');
     this.notice = document.querySelector('.notice');
+    this.logHeading = document.querySelector('.log-heading');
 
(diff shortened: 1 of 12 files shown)


src/lib/ext-monitor.js
```diff
@@ -81,6 +89,8 @@ export default class ExtensionMonitor {
     sendAllLogs: () => ({ existingLogs: this.logs }),
     loadLogs: (requestParams) => this.loadLogs(requestParams),
     getLoadedLogs: (requestParams) => this.getLoadedLogs(requestParams),
+    devToolsPanelShown: () => (this.isDevToolPanelOpen = true),
+    devToolsPanelHidden: () => (this.isDevToolPanelOpen = false),
```
rpl, line 93: This part seems to not be taking into account that we may have multiple extension devtools panels opened at the same time (as I described in this comment https://github.com/mozilla/extension-activity-monitor/issues/22#issuecomment-667996573), and so when one of the extensions devtools panels is hidden this is going to set `this.isDevToolsPanelOpen` to `false` even if another devtools panel has not been hidden.

would you mind to describe me what is the issue and how you are trying to solve it at the moment?

src/lib/ext-monitor.js
```diff
@@ -17,9 +19,15 @@ export default class ExtensionMonitor {
...
 
   async isActivityLogPageOpen() {
+    const activitylogPage = getActivityLogPageURL();
+    if (this.isDevToolPanelOpen) {
+      return true;
+    }
+
     const tab = await browser.tabs.query({
-      url: getActivityLogPageURL(),
+      url: [activitylogPage, `${activitylogPage}?*`],
```
rpl, line 28: at a first glance this looks unrelated to the main goal of this PR (the activitylog page running in the devtools panel), I guess that this is something you added while you were trying to use the activitylog page with the filterTabId parameter in a tab, is that right?

I guess that this would also make `isActivityLogPageOpen` to resolve to true when there are activitylog pages for the loaded log files opened in tabs, is that the case? is that expected?

src/lib/web-component/filter-timestamp/filter-timestamp-element.js
```diff
@@ -72,6 +72,16 @@ export class FilterTimestamp extends HTMLElement {
...
   connectedCallback() {
+    const searchParams = new URLSearchParams(
+      document.location.search.substring(1)
+    );
+
+    const filterTabId = searchParams.get('filterTabId');
+    // context menu doesn't work on devtool_page
+    if (filterTabId) {
+      return;
+    }
```
rpl, line 83: To disable the context menu while running in a devtools extension context we may use a different approach, e.g. one of the following ones should work:

- check if `browser.menus` and `browser.menus.onClicked` are defined, and early exit if they are not
- or check if `browser.devtools` is defined  

The first approach may be better, because it does runtime detection for the feature that are going to be needed for this component to be able to work as expected.

This way we can keep the "filter by timestamp" context menu is we open the activitylog page in the same mode (filtered by tab id) in an extension context where the context menu API is available.

Also, add a mention to the issue that is tracking the issue in the inline comment.

src/lib/web-component/log-view/log-view-element.js
```diff
@@ -115,8 +115,17 @@ export class LogView extends HTMLElement {
...
   connectedCallback() {
+    const searchParams = new URLSearchParams(
+      document.location.search.substring(1)
+    );
+
+    const filterTabId = searchParams.get('filterTabId');
+    // context menu doesn't work on devtool_page
+    if (!filterTabId) {
+      this.tableBody.addEventListener('contextmenu', this);
+    }
```
rpl, line 126: same comment here.

src/lib/ext-activitylog.js
```diff
@@ -251,6 +264,23 @@ class Controller {
       if (existingLogs.length) {
         this.handleNewLogs(existingLogs);
       }
+
+      if (filterTabId) {
+        // currently menu container is hidden when activity log page is
+        // filtered by tab id
+        this.view.menuContainer.hidden = true;
```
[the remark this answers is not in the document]
atiqueahmedziad: As the menu container contains save and clear logs button, I am hiding those. 
Since in the devtools panel, the user will see the filtered logs by tabId but when he presses the "save logs" button, it will save all the logs. So, having that button displayed may mislead the user. Same scenario with "clear logs" button too (As it also clears all the logs, instead of what is displayed in the devtools panel).

I think it will be a good time to enable these buttons after we implement [#27 :Save logs with/without filter applied](https://github.com/mozilla/extension-activity-monitor/issues/27). 

What do you think? Should we keep these buttons enabled here?

src/lib/ext-monitor.js
```diff
@@ -8,6 +8,9 @@ export default class ExtensionMonitor {
   // Map<string, Function>
   extensionMapList = new Map([]);
 
+  // Set<number>
+  devToolsPanelTabIds = new Set();
```
atiqueahmedziad, line 12: This contains the **active** devtools panels tabIds.

src/lib/ext-monitor.js
```diff
@@ -20,14 +23,15 @@ export default class ExtensionMonitor {
...
+
     return tab.length > 0;
   }
 
   // If the Activity Log page is open, each logs will be send to
   // Activity Log page (as soon as it is encountered).
   async sendLogs(details) {
     const isExtPageOpen = await this.isActivityLogPageOpen();
-    if (isExtPageOpen) {
+    if (isExtPageOpen || this.devToolsPanelTabIds.size) {
```
atiqueahmedziad, line 34: If there are any active ActivityLog devtools panel, the realtime logs will be sent.

src/lib/ext-monitor.js
```diff
@@ -81,6 +85,8 @@ export default class ExtensionMonitor {
     sendAllLogs: () => ({ existingLogs: this.logs }),
     loadLogs: (requestParams) => this.loadLogs(requestParams),
     getLoadedLogs: (requestParams) => this.getLoadedLogs(requestParams),
+    addPanelTabId: ({ tabId }) => this.devToolsPanelTabIds.add(tabId),
+    deletePanelTabId: ({ tabId }) => this.devToolsPanelTabIds.delete(tabId),
```
atiqueahmedziad, line 89: The `addPanelTabId` and `deletePanelTabId` acts when an ActivityLog Devtools panel is shown or hidden respectively.

src/lib/ext-activitylog.js
```diff
@@ -95,6 +99,18 @@ class Model {
     return true;
   }
 
+  matchFilterTabId({ tabId }) {
+    if (!this.filter.tabId) {
```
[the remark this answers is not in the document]
atiqueahmedziad: done. Thanks :-)

src/lib/ext-activitylog.js
```diff
@@ -95,6 +99,18 @@ class Model {
...
+      return true;
+    }
+
+    if (tabId && tabId === this.filter.tabId) {
+      return true;
+    }
+
+    return false;
```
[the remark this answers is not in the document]

src/lib/ext-monitor.js
atiqueahmedziad: These were for keeping track of devtools panel. If there was any active devtools panel, we would be sending real-time logs there. 
Currently, I have applied a different approach to fix this, considering we have multiple extension devtools panels opened at the same time. I am storing the tabIds where activityLog devtools panel is shown and remvoing tabIds where activityLog devtools panel is hidden.

src/lib/ext-monitor.js
atiqueahmedziad: Yeah, initially, it was for the activitylog page with the filterTabId parameter in a tab. I have updated this part excluding this change.

atiqueahmedziad: I think we should keep this change, but fixing it by:
```
[activitylogPage, `${activitylogPage}?filterTabId*`],
```
. this will allow us to open tab with any tabId and retrieve realtime logs of that tab.

src/lib/web-component/filter-timestamp/filter-timestamp-element.js
atiqueahmedziad: done. Thanks :-)

src/lib/web-component/log-view/log-view-element.js
atiqueahmedziad: fixed it :-)

src/lib/ext-monitor.js
atiqueahmedziad: updated :-)

src/lib/ext-activitylog.js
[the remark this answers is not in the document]
rpl: @atiqueahmedziad I would be ok to defer it to a follow up if we have reasons to, but this activitylog page should still make it clear to the user that the view is filtered by tabId, and at that point how much is this different from what would happen for the activitylog page that we open in a tab if the user did set some filters on it?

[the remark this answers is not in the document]
atiqueahmedziad: Yeah, the other filter options are visible and interact with the user (user has to click the checkbox to apply filter). But when we option the activitylog page in devtools, although user should know it filters the logs using current tab id, but the activitylog page's UI doesn't let the user know it has been filtered by tabId. 

I am adding a way to display in UI that activitylog page is filtered by tabId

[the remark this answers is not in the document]
atiqueahmedziad: @rpl I made it is visible that, logs are being filtered by tab id. And enabled the menus. See the screenshot. Is this okay?

Screenshot 2020-08-14 at 8 08 46 PM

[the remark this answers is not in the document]
atiqueahmedziad: @rpl we are unable to save logs from devtools panel since `downloads` API is unavailable. I am getting the following error - 
`can't access property "onChanged", browser.downloads is undefined`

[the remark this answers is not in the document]
Rob--W: Pass a message to the bg page and let the bg page call `browser.downloads.download`.

src/lib/ext-monitor.js
rpl: @atiqueahmedziad I'm not sure that the devtools panel's `onShown` and `onHidden` events are the right way of keeping track if the devtools panel does still exist (hidden doesn't mean that we have destroyed it).

Another approach to keep track of that would be to make the activitylog page to connect a port if it does want the background page to send real time messages, that port would be automatically disconnected if the extension page or extension panel has been destroyed and so we could use that to detect when we don't have any page listening for realtime logs anymore (we don't need to send the real time logs over the `runtime.Port`, we can keep broadcasting them using `browser.runtime.sendMessage` as we are doing at the moment, and we could be using the ports just to keep track of the extension contexts that are still alive)

The flow would likely be something like the following:

- the activitylog page detects that it does want to receive the realtime logs and so it calls `browser.runtime.connect({name: "monitor-realtime-logs"})`
- on the background page side, ext-monitor would be registering a listener on the `browser.runtime.onConnect` API event, and if the received port does have the expected name (`"monitor-realtime-logs"` in the example above) and if it does:
  - it keeps track of the port (e.g. in a `Set`)
  - it subscribe a listener on `port.onDisconnect`, and remove the port from the data structure where we keep track of them (e.g. from the `Set` I did mention before)
- ext-monitor will then be broadcasting the logs collected in real time if there are `runtime.Port` instances still alive (still part of the `Set` I did mention above)

How that sounds to you?

Some references to the APIs I did mention:
- https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/runtime/connect
- https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/runtime/onConnect
- https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/runtime/Port

atiqueahmedziad: Thanks @rpl for explaining in detail, I have implemented it :-)

src/lib/ext-activitylog.js
```diff
@@ -95,6 +99,14 @@ class Model {
     return true;
   }
 
+  matchFilterTabId({ tabId }) {
+    if (this.filter.tabId === null) {
```
rpl, line 103: `== null` to make sure this return earlier if tabId is `null` or `undefined`.

You may want to try this in a js console (e.g. a devtools webconsole panel or from a nodejs interactive interpreter) to better see the difference, eg. when using  `... == null`:

```
null == null // => true
undefined == null // => true
0 == null // false
```

versus when using `... === null`:

```
null === null // => true
undefined === null // => false
0 === null // => false
```

(but leave the `===` at line 107, there we want it to only return true on strict equality between the two values).

src/lib/ext-activitylog.js
```diff
@@ -251,6 +266,16 @@ class Controller {
       if (existingLogs.length) {
         this.handleNewLogs(existingLogs);
       }
+
+      if (filterTabId) {
+        this.view.logHeading.textContent = `${this.view.logHeading.textContent} Filtered By Tab Id: ${filterTabId}`;
```
rpl, line 271: As it is this looks a bit of a violation of the MVC pattern, the controller shouldn't change the textContent of the an element part by the view by directly accessing it as a property.

isn't the view already receiving the filterTabId as part of the filter properties got from the model?
if that is the case, then the view could use that to update also that header element accordingly.

src/lib/ext-monitor.js
```diff
@@ -102,6 +98,47 @@ export default class ExtensionMonitor {
     return logs;
   }
 
+  async saveLogs({ blob, filename }) {
+    const url = URL.createObjectURL(blob);
```
rpl, line 102: Aren't the log that we want to save already available in the `logs` property of the `ExtensionMonitor` class?

Sending a blob from the activitylog page seems unnecessary as it would send across IPC something that we do already have locally (and which can be potentially become pretty big if the monitoring session is long enough).

Even the filename doesn't seem to be necessary, given that the caller is sending a literal string value for it.

I think that this event handler could be written without taking any parameter for now (and we may add some, e.g. one with the filter options, once we actually need it).

src/lib/ext-activitylog.js
atiqueahmedziad: > isn't the view already receiving the filterTabId as part of the filter properties got from the model?
> if that is the case, then the view could use that to update also that header element accordingly.

The `log-view` web component is directly getting the filter object from model via `isFilterMatched` method. 
I am made a new method in new to deal with it. Does that work?

src/lib/ext-monitor.js
atiqueahmedziad: Updated it with empty parameter and using the logs from background.

src/lib/ext-activitylog.js
atiqueahmedziad: Thanks :-).  Updated with double equals :-)

src/lib/ext-activitylog.js
```diff
@@ -300,7 +331,7 @@ class Controller {
 
   async saveLogs() {
     try {
-      await save.saveAsJSON(this.model.logs);
+      await save.saveAsJSON();
```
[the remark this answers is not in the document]
atiqueahmedziad: Yeah, this is showing error whenever `saveLogs` throws error.

src/lib/ext-monitor.js
```diff
@@ -16,18 +19,10 @@ export default class ExtensionMonitor {
     });
   }
 
-  async isActivityLogPageOpen() {
-    const tab = await browser.tabs.query({
```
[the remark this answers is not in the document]
atiqueahmedziad: removed the `tabs` permission. Thanks for noticing.

src/lib/ext-activitylog.js
```diff
@@ -251,6 +270,18 @@ class Controller {
       if (existingLogs.length) {
         this.handleNewLogs(existingLogs);
       }
+
+      if (filterTabId) {
```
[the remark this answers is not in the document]
atiqueahmedziad: Filed issue #45

src/lib/ext-activitylog.js
```diff
@@ -184,6 +197,10 @@ class View {
   clearTable() {
     this.logView.clearTable();
   }
+
+  renderHeading({ filterTabId }) {
+    this.logHeading.textContent = `Activity Logs Filtered By Tab Id: ${filterTabId}`;
+  }
```
rpl, line 203: This is ok for now, at least it doesn't violate the MVC pattern as we discussed in some of the previous review comments, but after using the panel a big I think that as a user I would expect the "filter by tabId" to be visible along with the other filters (eventually only visible when the activitylog page has detected the filterTabId parameter as part of the url, and maybe without allowing the users to clear the filter as they can on the other filters).

Let's add it as a point in the agenda (so that we can agree if we want to file an issue for this and which priority should it get).

rpl:
@atiqueahmedziad I did just notice that this version does trigger a test failure on travis: https://travis-ci.org/github/mozilla/extension-activity-monitor/builds/718510182#L259

would you mind to take a look to it before merging this on master? it looks like it is just missing a mock and so it should be easy enough to track it down and quick to fix (but feel free to ping me if you think that you need another pair of eyes to evaluate it).

rpl:
@atiqueahmedziad thanks for fixing the test, feel free to merge this

PR #38, opened 2020-08-08
feat: start monitoring newly installed extensions automatically

### LINKED ISSUE #23: Listen for newly installed extensions and monitor them automatically
When the extension activity monitor is monitoring all the extensions and it is active, it should monitor for newly installed extensions and automatically monitor them.

atiqueahmedziad:
Fixes #23 

This PR does -
- Start monitor newly installed extensions automatically.
- Stop monitoring (removing listener) for uninstalled extensions automatically.

### DIFF
--- a/src/lib/ext-monitor.js
+++ b/src/lib/ext-monitor.js
@@ -9,9 +9,10 @@ export default class ExtensionMonitor {
   extensionMapList = new Map([]);
 
   async getAllExtensions() {
+    const selfId = browser.runtime.id;
     const extensions = await browser.management.getAll();
     return extensions.filter((extension) => {
-      return extension.type === 'extension' && extension.id !== this.selfId;
+      return extension.type === 'extension' && extension.id !== selfId;
     });
   }
 
@@ -86,7 +87,8 @@ export default class ExtensionMonitor {
   }
 
   onInstalledExtension = ({ id, type }) => {
-    if (type !== 'extension' || id === this.selfId) {
+    const selfId = browser.runtime.id;
+    if (type !== 'extension' || id === selfId) {
       return;
     }
 
@@ -158,7 +160,6 @@ export default class ExtensionMonitor {
   };
 
   init() {
-    this.selfId = browser.runtime.id;
     browser.runtime.onMessage.addListener(this.messageListener);
     browser.tabs.onRemoved.addListener(this.onRemovedListener);
   }
--- a/src/lib/ext-monitor.js
+++ b/src/lib/ext-monitor.js
@@ -9,10 +9,11 @@ export default class ExtensionMonitor {
   extensionMapList = new Map([]);
 
   async getAllExtensions() {
-    const selfId = browser.runtime.id;
     const extensions = await browser.management.getAll();
     return extensions.filter((extension) => {
-      return extension.type === 'extension' && extension.id !== selfId;
+      return (
+        extension.type === 'extension' && extension.id !== browser.runtime.id
+      );
     });
   }
 
@@ -87,8 +88,7 @@ export default class ExtensionMonitor {
   }
 
   onInstalledExtension = ({ id, type }) => {
-    const selfId = browser.runtime.id;
-    if (type !== 'extension' || id === selfId) {
+    if (type !== 'extension' || id === browser.runtime.id) {
       return;
     }
 
--- a/src/lib/ext-monitor.js
+++ b/src/lib/ext-monitor.js
@@ -10,9 +10,8 @@ export default class ExtensionMonitor {
 
   async getAllExtensions() {
     const extensions = await browser.management.getAll();
-    const self = await browser.management.getSelf();
     return extensions.filter((extension) => {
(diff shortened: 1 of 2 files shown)


src/lib/ext-monitor.js
```diff
@@ -9,7 +9,11 @@ export default class ExtensionMonitor {
     const extensions = await browser.management.getAll();
     const self = await browser.management.getSelf();
     return extensions.filter((extension) => {
-      return extension.type === 'extension' && extension.id !== self.id;
+      return (
+        extension.type === 'extension' &&
+        extension.id !== self.id &&
+        extension.enabled
```
Rob--W, line 15: Don't filter disabled extensions for now. Doing so prevents users from monitoring disabled extensions.

(and even if you did want to filter such events, then you would have to listen for `onEnabled` / `onDisabled` to detect changes in the "enabled" state).

src/lib/ext-monitor.js
```diff
@@ -97,7 +106,31 @@ export default class ExtensionMonitor {
     }
   };
 
+  onInstalledExtension = async (extInfo) => {
+    const { enabled, id, type } = extInfo;
+
+    const self = await browser.management.getSelf();
```
Rob--W, line 112: Use `browser.runtime.id` instead of `browser.management.getSelf().id`

(you can also use this simpler method in the `getAllExtensions` method)

src/lib/ext-monitor.js
```diff
@@ -97,7 +106,31 @@ export default class ExtensionMonitor {
...
+    const monitorStarted = this.hasActivityListeners();
```
Rob--W, line 113: It shouldn't be needed to check `hasActivityListeners`. The event is not fired when the listener has been unregistered.

src/lib/ext-monitor.js
```diff
@@ -10,9 +10,8 @@ export default class ExtensionMonitor {
 
   async getAllExtensions() {
     const extensions = await browser.management.getAll();
-    const self = await browser.management.getSelf();
     return extensions.filter((extension) => {
-      return extension.type === 'extension' && extension.id !== self.id;
+      return extension.type === 'extension' && extension.id !== this.selfId;
```
Rob--W, line 14: Is there a reason for storing the `browser.runtime.id` constant in a new member instead of directly using `browser.runtime.id`?

atiqueahmedziad: Not any strong reason, I just thought I am using the extension id two times, so, stored it in a variable and used the same variable twice. Now, I have updated it by using directly.

src/lib/ext-monitor.js
```diff
@@ -9,10 +9,10 @@ export default class ExtensionMonitor {
   extensionMapList = new Map([]);
 
   async getAllExtensions() {
+    const selfId = browser.runtime.id;
     const extensions = await browser.management.getAll();
-    const self = await browser.management.getSelf();
     return extensions.filter((extension) => {
-      return extension.type === 'extension' && extension.id !== self.id;
+      return extension.type === 'extension' && extension.id !== selfId;
```
Rob--W, line 15: Just use `browser.runtime.id` directly. There is no need for storing it in a temporary variable.

tests/ext-monitor.test.js
```diff
@@ -78,24 +94,85 @@ describe('start extension monitoring', () => {
...
+    expect(extMonitor.extensionMapList.size).toBe(0);
+
+    await extMonitor.startMonitor();
+    expect(extMonitor.extensionMapList.size).toBe(2);
+
+    // When a new extension is installed, `onInstalledExtension` method is
+    // being called with new extension's info.
+    extMonitor.onInstalledExtension(newExtension);
+
+    // Since this is a theme, it will not be monitored.
```
Rob--W, line 139: will -> should

tests/ext-monitor.test.js
```diff
@@ -78,24 +94,85 @@ describe('start extension monitoring', () => {
...
+        onInstalled: { addListener: onInstalledAddListener },
+        onUninstalled: { addListener: onUninstalledAddListener },
+      },
+      runtime: { id: selfExt.id },
+    };
+
+    const extMonitor = new ExtensionMonitor();
+    const getAllExtensionsFn = jest.spyOn(extMonitor, 'getAllExtensions');
+
+    extMonitor.selfId = selfExt.id;
```
Rob--W, line 126: `selfId` is not used any more. Why is it here?

src/lib/ext-monitor.js
atiqueahmedziad: done :-)

tests/ext-monitor.test.js
atiqueahmedziad: done :-)

tests/ext-monitor.test.js
atiqueahmedziad: sorry, didn't catch my eye. Thanks for noticing. Removed now :-)

PR #40, opened 2020-08-09
feat: Clear the start and stop timestamps filter separately

### LINKED ISSUE #33: Clear the start and stop timestamps separately
This is a follow-up from [PR#17](https://github.com/mozilla/extension-activity-monitor/pull/17#discussion_r466972692). 

Currently, when we clear the timestamp, it [clears the start and stop timestamps at once](https://github.com/mozilla/extension-activity-monitor/blob/master/src/lib/web-component/filter-timestamp/filter-timestamp-element.js#L66). 

We should be able to clear start and stop timestamp separately from UI.

atiqueahmedziad:
This pr fixes #33 
Now, we can clear start and stop timestamps filters separately.

### Screenshot
Screenshot 2020-08-10 at 2 28 12 AM


src/lib/ext-activitylog.js
```diff
@@ -9,7 +9,7 @@ class Model {
       type: new Set(),
       name: new Set(),
       keyword: '',
-      timeStamp: null,
+      timeStamp: {},
```
rpl, line 12: @atiqueahmedziad why you had to set this back to an empty object instead of null? at a first glance this change looks not strictly needed for this PR, would you mind to briefly explain the reason behind this choice?

src/lib/web-component/filter-timestamp/filter-timestamp-element.js
```diff
@@ -58,29 +58,62 @@ export class FilterTimestamp extends HTMLElement {
     );
   }
 
-  onClearFilter = () => {
-    this.startTimeLabel.textContent = 'From Beginning';
-    this.stopTimeLabel.textContent = 'Up to End';
-    this.filterContainer.hidden = true;
+  onClearFilter(clearRequest) {
```
rpl, line 61: Another option would be to make this function to take two boolean parameters, `clearStart` and `clearStop`.

src/lib/ext-activitylog.js
atiqueahmedziad: Previously, I made it `null` when we don't have any timestamp filters applied. To handle different types i.e. `object` and `null` in a single variable, I had to [add some extra codes](https://github.com/mozilla/extension-activity-monitor/pull/40/files#diff-0d42cddec555ff90f252f9e7b5f94496L31-L35) which I can avoid if I stick to one type i.e. `object` only. Also, I will have to put another condition [here](https://github.com/mozilla/extension-activity-monitor/pull/40/files#diff-0d42cddec555ff90f252f9e7b5f94496R76-R78) to check if timestamp is an empty object. If it's empty, I will assign `null` to it. And that will add some more extra codes. 

@rpl what do think about this? Should we do keep it `null`?

rpl: @atiqueahmedziad the link in your comment (https://github.com/mozilla/extension-activity-monitor/pull/40/files#diff-0d42cddec555ff90f252f9e7b5f94496L31-L35) isn't pointing me to that extra code, would you mind to paste me a snippet in a comment so that I can see if there was a way to make that simpler?

atiqueahmedziad: oh, sorry. I was pointing to the following code which is in L31-L35 in `ext-activitylog.js` file in this PR. 
```
    // this.timestamp is null when timestamp filter is not applied.
    if (!this.timeStamp) {
      this.timeStamp = {};
    }
```
I had to assign `this.timeStamp` to empty object and then I put value to `start` or `stop` property of the `this.timeStamp` object.

src/lib/web-component/filter-timestamp/filter-timestamp-element.js
```diff
@@ -28,18 +31,15 @@ export class FilterTimestamp extends HTMLElement {
       return;
     }
 
-    // this.timestamp is null when timestamp filter is not applied.
-    if (!this.timeStamp) {
-      this.timeStamp = {};
-    }
-
```
atiqueahmedziad, line 35: @rpl here is the part, I was talking about on the other comment. I can avoid this if I use a single type i.e. `object` only.

rpl: @atiqueahmedziad it doesn't seem a big deal, you could also do something like this:

```js
const timeStamp = this.timeStamp || {}

....
if (info.menuItemId === 'startTime') {
  timeStamp.start = ...
  ...
} else if (info.menuItemId === 'stopTime') {
  timeStamp.stop = ...
  ...
}

...
this.timeStamp = timeStamp;
this.dispatchFilterChange();
```

and avoid the check `Object.keys(this.filter.timeStamp).length === 0`, which I personally don't like that much.

does this sounds a reasonable alternative for the issue you had?

src/lib/web-component/filter-timestamp/filter-timestamp-element.js
atiqueahmedziad: I have added boolean parameters now.

src/lib/web-component/filter-timestamp/filter-timestamp-element.js
atiqueahmedziad: yes, this sounds good. But I had to use `Object.keys(this.filter.timeStamp).length === 0` while assigning empty object to null [here](https://github.com/mozilla/extension-activity-monitor/pull/40/commits/19f9545d2416104c5d450cbdb632b20328d06c55#diff-0d42cddec555ff90f252f9e7b5f94496R84).

src/lib/web-component/filter-timestamp/filter-timestamp-element.js
```diff
@@ -58,29 +62,70 @@ export class FilterTimestamp extends HTMLElement {
...
+
+    if (!clearStart && !clearStop) {
+      this.startTimeLabel.textContent = 'From Beginning';
+      this.stopTimeLabel.textContent = 'Up to End';
+      this.timeStamp = {};
+    }
+
+    if (Object.keys(this.timeStamp).length === 0) {
+      this.timeStamp = null;
+    }
```
atiqueahmedziad, line 86: @rpl I had to use `Object.keys(this.timeStamp).length === 0` here.

rpl: @atiqueahmedziad That's fine, I don't like it but it is not such a big deal.

But please be consistent with the time passed to the parameters, if they are supposed to be boolean, pass a boolean for them or omit them completely and set a default value for the variable.

e.g instead of
```
      switch (event.target) {
        case this.clearFilterBtn:
          this.onClearFilter(null, null);
          break;
        case this.clearStartTimeBtn:
          this.onClearFilter(true, null);
          break;
        case this.clearStopTimeBtn:
          this.onClearFilter(null, true);
          break;
      }
```

we could define this function as:

```
  onClearFilter(clearStart = true, clearStop = true) {
    ...
  }
```

and then call it without any parameter if we want to clear both start and stop, and with two explicit boolean parameters if we want to clear only one of them:

```
      switch (event.target) {
        case this.clearFilterBtn:
          this.onClearFilter();
          break;
        case this.clearStartTimeBtn:
          this.onClearFilter(true, false);
          break;
        case this.clearStopTimeBtn:
          this.onClearFilter(false, true);
          break;
      }
```

src/lib/web-component/filter-timestamp/filter-timestamp-element.js
```diff
@@ -58,29 +62,70 @@ export class FilterTimestamp extends HTMLElement {
...
+      this.stopTimeLabel.textContent = 'Up to End';
+      this.clearStopTimeBtn.hidden = true;
+      delete this.timeStamp.stop;
+    }
+
+    if (!clearStart && !clearStop) {
+      this.startTimeLabel.textContent = 'From Beginning';
+      this.stopTimeLabel.textContent = 'Up to End';
+      this.timeStamp = {};
+    }
```
rpl, line 82: @atiqueahmedziad at a first glance I would say that this shouldn't be needed, in which case we would need this along with the other two at lines 66-69 and lines 72-76?

src/lib/web-component/filter-timestamp/filter-timestamp.css
```diff
@@ -37,4 +49,9 @@
 .start-time-container,
 .stop-time-container {
   display: flex;
+  align-items: center;
+}
+
+[hidden] {
+  display: none;
 }
```
rpl, line 57: @atiqueahmedziad this shouldn't be actually needed, did you had any actual issue that you are workarounding with this?

atiqueahmedziad: The close icon along with start and stop timestamp does not disappear when until I set this. Below, I shared a screenshot, when I didn't set this CSS.
### screenshot 
Screenshot 2020-08-13 at 12 28 54 AM

src/lib/web-component/filter-timestamp/filter-timestamp-element.js
atiqueahmedziad: Yes, Thanks for the idea. :-)

src/lib/web-component/filter-timestamp/filter-timestamp.css
rpl: @atiqueahmedziad I see, yeah that seems to be because the "hidden attribute is not being applied to the elements inside the shadow DOM" (/me sigh for a couple of seconds :-)), would you mind to add an inline comment right above it to explicitly mention that reason?

src/lib/web-component/filter-timestamp/filter-timestamp-element.js
```diff
@@ -58,29 +62,64 @@ export class FilterTimestamp extends HTMLElement {
...
+  }
+
   connectedCallback() {
     browser.menus.onClicked.addListener(this.setFilterRange);
     browser.menus.onHidden.addListener(this.onHiddenListener);
-    this.clearBtn.addEventListener('click', this.onClearFilter);
+    this.clearFilterBtn.addEventListener('click', this);
+    this.clearStartTimeBtn.addEventListener('click', this);
+    this.clearStopTimeBtn.addEventListener('click', this);
```
rpl, line 114: you should be able to subscribe a single click event listener (e.g. on `this.filterContainer`) without any additional changes (the `handleEvent` is already conditioned on the `event.target` property).

atiqueahmedziad: done :-) Thanks.

src/lib/web-component/filter-timestamp/filter-timestamp.css
atiqueahmedziad: Yes, I have added :-)

src/lib/web-component/filter-timestamp/filter-timestamp-element.js
```diff
@@ -58,29 +62,62 @@ export class FilterTimestamp extends HTMLElement {
...
+          this.onClearFilter();
+          break;
+        case this.clearStartTimeBtn:
+          this.onClearFilter(true, false);
+          break;
+        case this.clearStopTimeBtn:
+          this.onClearFilter(false, true);
+          break;
+        default:
+          break;
```
rpl, line 106: The `default` in this `switch...case` is not really needed, this method would already turns into a no-op for events that we don't want to handle (because they are not coming from one of the event.target that we expect).

atiqueahmedziad: done :-)

PR #46, opened 2020-08-17
fix: Made timestamp consistant

### LINKED ISSUE #41: timestamp should be consistant
As [mentioned on PR #28](https://github.com/mozilla/extension-activity-monitor/pull/28#issuecomment-671541936), the timestamps should be consistent in any format of log view. Currently, in a saved log file, the timestamp has a different format compared to the timestamps in activity log page.
## Screenshot

diff_between_timestamps_in_realtime_vs_loadedlogs

atiqueahmedziad:
Fixes #41 

- `log.timeStamp` contains timestamp (number).
- While viewing logs, the timestamp is being formatted into human-readable format i.e. `8:49:31 PM` (`hh:mm:ss am/pm`)
- When hover the timestamp in `log-view`, it shows tooltip in the format `17-8-2020 8:49:31 PM` 
- The timestamp filter texts also show the same format as tooltip i.e. `17-8-2020 8:49:31 PM` 

## Screenshot
Screenshot 2020-08-17 at 8 52 49 PM


src/lib/web-component/filter-timestamp/filter-timestamp-element.js
```diff
@@ -25,22 +27,26 @@ export class FilterTimestamp extends HTMLElement {
 
   setFilterRange = (info) => {
     const selectedEl = browser.menus.getTargetElement(info.targetElementId);
-    const selectedRow = selectedEl.closest('tr');
+    if (!selectedEl) {
+      return;
+    }
```
atiqueahmedziad, line 32: When we have multiple activity log page opened at a time, multiple context menu event listeners are being assigned but we get only one `selectedEl` (where the context event has been called)

src/lib/ext-listen.js
```diff
@@ -31,3 +31,14 @@ export function openActivityLogPage() {
...
+
+export function timeFormat(timestamp) {
+  const date = new Date(timestamp);
+  return `${date.toLocaleTimeString()}`;
+}
+
+export function dateTimeFormat(timestamp) {
+  const date = new Date(timestamp);
+  const month = date.getMonth() + 1;
+  return `${date.getDate()}-${month}-${date.getFullYear()} ${date.toLocaleTimeString()}`;
```
Rob--W, line 43: `dd-MM-YYYY` looks reasonable to me because the format is the same as where I am from (The Netherlands), but it is not universally true.

Since you're trying to have locale-aware time strings, I suggest to be consistent.

Use the `Intl.DateTimeFormat` API to format the date in a locale-aware way - https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/DateTimeFormat

src/lib/web-component/filter-timestamp/filter-timestamp-element.js
```diff
@@ -33,14 +35,15 @@ export class FilterTimestamp extends HTMLElement {
 
     const timeStamp = this.timeStamp || {};
 
-    const chosenTimestamp = selectedRow.querySelector('.timestamp').textContent;
+    const chosenTimestamp = selectedRow.querySelector('.timestamp')?._timestamp;
```
Rob--W, line 38: Use `selectedRow._log.timeStamp` instead.

I would rather not have an external class using an internal `_log` property, but it is preferable to your current approach, of adding a `_timeStamp` property to an unrelated DOM element.

To reduce the risk of future regressions, it would be nice to have a unit test that verifies that the time labels are rendered as expected.

src/lib/ext-listen.js
```diff
@@ -31,3 +31,19 @@ export function openActivityLogPage() {
     url: getActivityLogPageURL(),
   });
 }
+
+export function dateTimeFormat(timestamp, options) {
+  const dateTime = new Date(timestamp);
+  const time = dateTime.toLocaleTimeString();
```
Rob--W, line 37: `DateTimeFormat` also supports parameters to include the time in the generated string. Do not use `toLocaleTimeString`, but only `DateTimeFormat`.

tests/ext-activitylog.test.js
```diff
@@ -269,3 +270,77 @@ test('clearing logs from activitylog page', async () => {
...
+      connect,
+    },
+  };
+
+  sendMessage.mockImplementation(() => {
+    return Promise.resolve({ existingLogs: [] });
+  });
+
+  document.body.innerHTML = activityLogBody;
+  const dateTimeFormatFn = jest.spyOn(ExtListen, 'dateTimeFormat');
```
Rob--W, line 305: You shouldn't mock your function, but the API that you're using. The only changes should be:
- fix `'en-US'` as the locale instead of `undefined`.
- fix the `timeZone` option.

By doing both, the generated time strings become fully deterministic / predictable, and you can put the expected times in strings below.

tests/ext-activitylog.test.js
```diff
@@ -269,3 +270,77 @@ test('clearing logs from activitylog page', async () => {
...
+
+    return `${expectedDate} ${time}`;
+  });
+
+  activityLog.handleNewLogs(logs);
+
+  expect(dateTimeFormatFn).toHaveBeenCalled();
+
+  const dateTime = new Date(logTimestamp);
+  const expectedTime = dateTime.toLocaleTimeString();
```
Rob--W, line 336: Hard-code the expected strings instead of generating it based on the dynamic objects.

src/lib/ext-listen.js
```diff
@@ -31,3 +31,19 @@ export function openActivityLogPage() {
     url: getActivityLogPageURL(),
   });
 }
+
+export function dateTimeFormat(timestamp, options) {
```
Rob--W, line 35: Why did you put this function in `ext-listen.js` instead of a new file called, say, `formatters.js`?

PR #48, opened 2020-08-19
"undefined" viewType displayed as 'other' in filter checkbox

### LINKED ISSUE #42: ViewTypes filter options should not contain undefined
While working with PR #31, I noticed for `log.type` = `api_call` the `viewType` can be `undefined`. See the screenshot below.

Currently we are only [filtering with `log.type` = `content_script`](https://github.com/mozilla/extension-activity-monitor/blob/master/src/lib/ext-activitylog.js#L166) to exclude the `undefined` from being added as a `viewType` filter option. I think we should be filtering out with `log.viewType` = `undefined` instead of `log.type` = `content_script` as `viewType` can be `undefined` in both log.type = content_scirpt and log.type = api_call.

## Screenshot
Screenshot 2020-08-12 at 9 49 50 PM

atiqueahmedziad:
Fixes #42

### DIFF
--- a/src/lib/web-component/filter-option/filter-option-element.js
+++ b/src/lib/web-component/filter-option/filter-option-element.js
@@ -5,6 +5,8 @@ export class FilterOption extends HTMLElement {
     this.viewCheckboxLabels = new Set();
     // It contains the checked checkboxes
     this.activeCheckboxLabels = new Set();
+    // When viewtype is undefined for API calls
+    this.FILTER_OPTION_UNDEFINED_LABEL = 'other';
 
     const shadow = this.attachShadow({ mode: 'open' });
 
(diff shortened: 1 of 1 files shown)


src/lib/web-component/filter-option/filter-option-element.js
```diff
@@ -42,10 +42,10 @@ export class FilterOption extends HTMLElement {
   addNewCheckbox(checkboxLabel) {
     const newCheckbox = this.checkboxTemplate.cloneNode(true);
     newCheckbox.querySelector('label span').textContent =
-      checkboxLabel || 'undefined';
+      checkboxLabel || 'other';
```
Rob--W, line 45: Rather than duplicating the same string thrice, put it in a constant with a good name and use that instead.

atiqueahmedziad: Added `otherLabel = `other'` at constructor and used this. :-)

src/lib/web-component/filter-option/filter-option-element.js
```diff
@@ -5,6 +5,7 @@ export class FilterOption extends HTMLElement {
     this.viewCheckboxLabels = new Set();
     // It contains the checked checkboxes
     this.activeCheckboxLabels = new Set();
+    this.otherLabel = 'other';
```
Rob--W, line 8: To emphasize that this is an immutable value, use `const FILTER_OPTION_UNDEFINED_LABEL = 'other';` (or a shorter name), with a brief comment before it to explain what it's used for.

atiqueahmedziad: Would I declare it in constructor? Or declare it as `const` inside the methods?

Rob--W: Outside the class.

atiqueahmedziad: updated :-)

Rob--W (approved):

PR #49, opened 2020-08-21
feat: restoring applied filters with search params

### LINKED ISSUE #24: Restore filters on page load
Currently, the filters are lost once we reload the activitylog page or if we accidentally close the activitylog page. But we want the filters to be applied (that we set previously) on activitylog page loads.

atiqueahmedziad:
Fixes #24 

This PR updates URL with filter information as soon as any filter is applied/ removes. This enables the user to restore the filters via URL.

### DIFF
--- a/src/devtools/devtools.js
+++ b/src/devtools/devtools.js
@@ -3,5 +3,5 @@ const tabId = browser.devtools.inspectedWindow.tabId;
 browser.devtools.panels.create(
   'Extension Activity',
   '/icons/eam-48.png',
-  `/activitylog/activitylog.html?filterTabId=${tabId}`
+  `/activitylog/activitylog.html?tabId=${tabId}`
 );
--- a/src/lib/ext-activitylog.js
+++ b/src/lib/ext-activitylog.js
@@ -185,7 +185,15 @@ class View {
   }
 
   setExistedFilters(updateFilter) {
-    const { id, viewType, type, name, keyword, timeStamp } = updateFilter;
+    const {
+      id,
+      viewType,
+      type,
+      name,
+      keyword,
+      timeStamp,
+      tabId,
+    } = updateFilter;
 
     this.extFilter.setExistedFilter(id);
     this.viewTypeFilter.setExistedFilter(viewType);
@@ -193,6 +201,15 @@ class View {
     this.apiNameFilter.setExistedFilter(name);
     this.keywordFilter.setExistedFilter(keyword);
     this.timestampFilter.setExistedFilter(timeStamp);
+
+    if (tabId) {
+      this.renderHeading({ tabId });
+      const filterDetail = { updateFilter: { tabId } };
+
+      this.logHeading.dispatchEvent(
+        new CustomEvent('filterchange', { detail: filterDetail })
+      );
+    }
   }
 
   setError(errorMessage) {
@@ -209,8 +226,8 @@ class View {
     this.logView.clearTable();
   }
 
-  renderHeading({ filteredTabId }) {
-    this.logHeading.textContent = `Activity Logs Filtered By Tab Id: ${filteredTabId}`;
+  renderHeading({ tabId }) {
+    this.logHeading.textContent = `Activity Logs Filtered By Tab Id: ${tabId}`;
   }
 }
 
@@ -229,6 +246,7 @@ class Controller {
     this.view.apiTypeFilter.addEventListener('filterchange', this);
     this.view.apiNameFilter.addEventListener('filterchange', this);
     this.view.timestampFilter.addEventListener('filterchange', this);
+    this.view.logHeading.addEventListener('filterchange', this);
 
     const searchParams = new URLSearchParams(
       document.location.search.substring(1)
@@ -286,18 +304,6 @@ class Controller {
 
       if (existingLogs.length) {
         this.handleNewLogs(existingLogs);
-      }
-
-      if (this.model.filter.tabId != null) {
-        const filteredTabId = this.model.filter.tabId;
-
-        this.view.renderHeading({ filteredTabId });
-
-        const filterDetail = {
-          updateFilter: { tabId: filteredTabId },
-        };
-
-        this.onFilterChange(filterDetail);
       }
     }
   }
--- a/src/lib/ext-activitylog.js
+++ b/src/lib/ext-activitylog.js
@@ -235,13 +235,7 @@ class Controller {
     );
 
     const loadedFileName = searchParams.get('file');
-    const filterTabId = searchParams.get('filterTabId');
-
-    if (document.location.search) {
-      const updateFilter = deSerializeFilters(searchParams);
-      this.model.setFilter(updateFilter);
-      this.view.setFilterOptionsFromURL(updateFilter);
-    }
+    const filterTabId = parseInt(searchParams.get('filterTabId'), 10);
 
     if (loadedFileName) {
       this.view.menuContainer.hidden = true;
@@ -282,6 +276,12 @@ class Controller {
           throw new Error(`wrong request type found - ${requestType}`);
         }
       });
+
+      if (document.location.search) {
+        const updateFilter = deSerializeFilters(searchParams);
+        this.model.setFilter(updateFilter);
+        this.view.setFilterOptionsFromURL(updateFilter);
+      }
 
       const existingLogs = await this.getExistingLogs();
 
--- a/src/lib/ext-activitylog.js
+++ b/src/lib/ext-activitylog.js
@@ -183,6 +183,17 @@ class View {
     this.apiNameFilter.updateFilterCheckboxes(filteredLogs);
   }
 
+  setFilterOptionsFromURL(updateFilter) {
+    const { id, viewType, type, name, keyword, timeStamp } = updateFilter;
+
+    this.extFilter.setFilterFromURL(id);
+    this.viewTypeFilter.setFilterFromURL(viewType);
+    this.apiTypeFilter.setFilterFromURL(type);
+    this.apiNameFilter.setFilterFromURL(name);
+    this.keywordFilter.setFilterFromURL(keyword);
+    this.timestampFilter.setFilterFromURL(timeStamp);
+  }
+
   setError(errorMessage) {
     if (errorMessage) {
       this.notice.textContent = errorMessage;
@@ -221,12 +232,20 @@ class Controller {
     const searchParams = new URLSearchParams(
       document.location.search.substring(1)
     );
-    const fileName = searchParams.get('file');
-    const filterTabId = parseInt(searchParams.get('filterTabId'), 10);
 
-    if (fileName) {
+    const loadedFileName = searchParams.get('file');
+
+    const filterIds = searchParams.get('ids');
+    const filterViewTypes = searchParams.get('viewTypes');
+    const filterTypes = searchParams.get('types');
+    const filterNames = searchParams.get('names');
+    const filterTabId = parseInt(searchParams.get('filterTabId'), 10);
+    const filterKeyword = searchParams.get('keyword');
+    const filterTimestamp = searchParams.get('timestamp');
+
+    if (loadedFileName) {
       this.view.menuContainer.hidden = true;
-      document.title = `Loaded Logs - ${fileName}`;
+      document.title = `Loaded Logs - ${loadedFileName}`;
 
       const currentTab = await browser.tabs.getCurrent();
       let logs;
@@ -263,6 +282,37 @@ class Controller {
           throw new Error(`wrong request type found - ${requestType}`);
         }
       });
+
+      const ids = filterIds?.split(',');
+
+      const viewTypes = filterViewTypes?.split(',');
+      // converting empty string to undefined
+      if (viewTypes?.length) {
+        const undefinedIndex = viewTypes.indexOf('');
+        if (undefinedIndex !== -1) {
+          viewTypes[undefinedIndex] = undefined;
+        }
+      }
+
+      const types = filterTypes?.split(',');
+      const names = filterNames?.split(',');
+      const timestamps = filterTimestamp?.split(',').map((timestamp) => {
+        return isNaN(timestamp) ? null : Number(timestamp);
+      });
+
+      // when we open activitylog.html having search parameters
+      if (document.location.search) {
+        const updateFilter = {
+          id: new Set(ids),
+          viewType: new Set(viewTypes),
+          type: new Set(types),
+          name: new Set(names),
+          keyword: filterKeyword,
+          timeStamp: timestamps,
+        };
+
+        this.view.setFilterOptionsFromURL(updateFilter);
+      }
 
       const existingLogs = await this.getExistingLogs();
 
@@ -336,14 +386,32 @@ class Controller {
   }
 
   onFilterChange(filterDetail) {
-    const { updateFilter, isNewFilterAdded } = filterDetail;
+    const { updateFilter } = filterDetail;
 
     this.model.setFilter(updateFilter);
-    // When new filter checkbox is added, it is in checked condition by default
-    // No need to re-render the rows then.
-    if (!isNewFilterAdded) {
-      this.view.setLogFilter((log) => this.isFilterMatched(log));
-    }
+    this.setFilterSearchParams(this.model.filter);
+    this.view.setLogFilter((log) => this.isFilterMatched(log));
+  }
+
+  setFilterSearchParams(filterObj) {
+    const { id, viewType, type, name, tabId, keyword, timeStamp } = filterObj;
+
+    const startTime = timeStamp?.start ? timeStamp.start : null;
+    const stopTime = timeStamp?.stop ? timeStamp.stop : null;
+
+    const idsQuery = `ids=${[...id].toString()}`;
+    const viewTypesQuery = `viewTypes=${[...viewType].toString()}`;
+    const typesQuery = `types=${[...type].toString()}`;
+    const namesQuery = `names=${[...name].toString()}`;
+    const timestampsQuery = `timestamp=${startTime},${stopTime}`;
+    const keywordQuery = `keyword=${keyword}`;
+    const tabIdQuery = `filterTabId=${tabId}`;
+
+    history.replaceState(
+      null,
+      null,
+      `?${idsQuery}&${viewTypesQuery}&${typesQuery}&${namesQuery}&${keywordQuery}&${tabIdQuery}&${timestampsQuery}`
+    );
   }
 
   isFilterMatched(log) {
--- a/src/lib/ext-activitylog.js
+++ b/src/lib/ext-activitylog.js
@@ -1,13 +1,14 @@
 import { save } from './save-load.js';
+import { serializeFilters, deSerializeFilters } from './formatters.js';
 
 class Model {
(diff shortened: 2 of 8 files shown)


src/lib/ext-activitylog.js
```diff
@@ -264,6 +283,37 @@ class Controller {
         }
       });
 
+      const ids = filterIds?.split(',');
```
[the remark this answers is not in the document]
atiqueahmedziad: This is done, I have used `JSON.stringify` and `JSON.parse` :-)

src/lib/ext-activitylog.js
```diff
@@ -264,6 +283,37 @@ class Controller {
...
+
+      const viewTypes = filterViewTypes?.split(',');
+      // converting empty string to undefined
```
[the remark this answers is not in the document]
atiqueahmedziad: Thanks :-) Done.

src/lib/ext-activitylog.js
```diff
@@ -221,12 +232,20 @@ class Controller {
...
+    const filterIds = searchParams.get('ids');
+    const filterViewTypes = searchParams.get('viewTypes');
+    const filterTypes = searchParams.get('types');
+    const filterNames = searchParams.get('names');
     const filterTabId = parseInt(searchParams.get('filterTabId'), 10);
+    const filterKeyword = searchParams.get('keyword');
+    const filterTimestamp = searchParams.get('timestamp');
 
-    if (fileName) {
+    if (loadedFileName) {
```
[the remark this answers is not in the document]
atiqueahmedziad: yes :-) Fixed it.

src/lib/ext-activitylog.js
```diff
@@ -336,14 +386,32 @@ class Controller {
...
+    const typesQuery = `types=${[...type].toString()}`;
+    const namesQuery = `names=${[...name].toString()}`;
+    const timestampsQuery = `timestamp=${startTime},${stopTime}`;
+    const keywordQuery = `keyword=${keyword}`;
+    const tabIdQuery = `filterTabId=${tabId}`;
+
+    history.replaceState(
+      null,
+      null,
+      `?${idsQuery}&${viewTypesQuery}&${typesQuery}&${namesQuery}&${keywordQuery}&${tabIdQuery}&${timestampsQuery}`
```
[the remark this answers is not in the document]
atiqueahmedziad: Used `URLSearchParams` and moved the serialization and de-serialization logic in `formatters.js`

src/lib/web-component/filter-option/filter-option-element.js
```diff
@@ -35,11 +37,21 @@ export class FilterOption extends HTMLElement {
...
+
+    this.dispatchFilterChangeEvent();
+
+    // When any new unknown filter option is found, it is enabled by default.
+    this.isCheckboxChecked = () => true;
+  }
+
+  setFilterFromURL(searchParamLabels) {
+    if (searchParamLabels.size > 0) {
+      this.isCheckboxChecked = (checkboxLabel) =>
```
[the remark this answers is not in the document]
atiqueahmedziad: done :-) Currently I am storing the unchecked filter options in `filter` object.

src/lib/ext-activitylog.js
```diff
@@ -264,6 +283,37 @@ class Controller {
...
+        const updateFilter = {
+          id: new Set(ids),
+          viewType: new Set(viewTypes),
+          type: new Set(types),
+          name: new Set(names),
+          keyword: filterKeyword,
+          timeStamp: timestamps,
+        };
+
+        this.view.setFilterOptionsFromURL(updateFilter);
```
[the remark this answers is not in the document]
atiqueahmedziad: Now, the flow is URL params -> Controller -> Model -> View ...

Rob--W (commented):
Review not complete yet, but submitting so we can discuss it during the current meeting.

src/lib/web-component/filter-option/filter-option-element.js
```diff
@@ -6,8 +6,8 @@ export class FilterOption extends HTMLElement {
     super();
 
     this.viewCheckboxLabels = new Set();
-    // It contains the checked checkboxes
-    this.activeCheckboxLabels = new Set();
+    // It contains the unchecked checkboxes
+    this.inActiveCheckboxLabels = new Set();
```
Rob--W, line 10: Rename to `inactive`, `inActiveCheckboxLabels` looks like "in active checkbox labels", which doesn't reflect the actual purpose.

src/lib/ext-activitylog.js
```diff
@@ -183,6 +184,17 @@ class View {
     this.apiNameFilter.updateFilterCheckboxes(filteredLogs);
   }
 
+  setFilterOptionsFromURL(updateFilter) {
+    const { id, viewType, type, name, keyword, timeStamp } = updateFilter;
+
+    this.extFilter.setFilterFromURL(id);
```
Rob--W, line 190: `setFilterOptionsFromURL` and `setFilterFromURL` are independent of the URL. For example, if we ever store the filter in a file in the future, the same logic can be used. Please choose a meaningful than that is generic enough to serve the current use case and this potential future use case.

Rob--W (commented):
After following the previous suggestion of making the `serialize` / `deserialize` methods stateless / independent of external state, it should be much easier to write unit tests to verify that those two methods work as intended. Please include these tests in the PR, and try to think of edge cases (e.g. incorrect/missing input).

src/devtools/devtools.js
```diff
@@ -3,5 +3,5 @@ const tabId = browser.devtools.inspectedWindow.tabId;
 browser.devtools.panels.create(
   'Extension Activity',
   '/icons/eam-48.png',
-  `/activitylog/activitylog.html?filterTabId=${tabId}`
+  `/activitylog/activitylog.html?tabId=${tabId}`
```
atiqueahmedziad, line 6: To make it consistent with the `filter` object property i.e. `tabId`

src/lib/ext-activitylog.js
atiqueahmedziad: I renamed it to `setExistedFilters` since these filters were set before.

src/lib/web-component/filter-option/filter-option-element.js
atiqueahmedziad: done :-)

src/lib/ext-activitylog.js
```diff
@@ -221,12 +233,13 @@ class Controller {
     const searchParams = new URLSearchParams(
       document.location.search.substring(1)
     );
-    const fileName = searchParams.get('file');
+
+    const loadedFileName = searchParams.get('file');
     const filterTabId = parseInt(searchParams.get('filterTabId'), 10);
```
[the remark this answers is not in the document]

src/lib/formatters.js
```diff
@@ -9,3 +9,69 @@ export function dateTimeFormat(timestamp, options) {
...
+
+    if (
+      key === 'id' ||
+      key === 'viewType' ||
+      key === 'type' ||
+      key === 'name'
+    ) {
+      searchParamVal = JSON.stringify([...updateFilter[key].exclude]);
+    } else if (key === 'timeStamp') {
+      searchParamVal = timestampFormat(updateFilter[key]);
```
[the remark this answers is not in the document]
atiqueahmedziad: Sorry, Fixed it :-)

src/lib/formatters.js
```diff
@@ -9,3 +9,69 @@ export function dateTimeFormat(timestamp, options) {
...
+
+function timestampFormat(timeStamp) {
+  const startTime = timeStamp?.start ? timeStamp.start : null;
+  const stopTime = timeStamp?.stop ? timeStamp.stop : null;
+  return JSON.stringify([JSON.stringify(startTime), JSON.stringify(stopTime)]);
+}
+
+export function serializeFilters(updateFilter) {
+  const currentURL = new URL(document.location.href);
+  const searchParams = new URLSearchParams(currentURL.search.slice(1));
```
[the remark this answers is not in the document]
atiqueahmedziad: Yes done now :-)

src/lib/formatters.js
```diff
@@ -9,3 +9,69 @@ export function dateTimeFormat(timestamp, options) {
...
+    } else if (key === 'timeStamp') {
+      searchParamVal = timestampFormat(updateFilter[key]);
+    } else {
+      searchParamVal = updateFilter[key];
+    }
+
+    searchParams.set(key, searchParamVal);
+  });
+
+  history.replaceState(null, null, `?${searchParams}`);
```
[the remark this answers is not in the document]
atiqueahmedziad: done :-)

src/lib/ext-activitylog.js
```diff
@@ -264,17 +276,25 @@ class Controller {
...
+
       const existingLogs = await this.getExistingLogs();
 
       if (existingLogs.length) {
         this.handleNewLogs(existingLogs);
       }
 
-      if (filterTabId) {
-        this.view.renderHeading({ filterTabId });
+      if (this.model.filter.tabId != null) {
```
Rob--W, line 291: This logic would better be placed in `setExistedFilters`.

src/lib/formatters.js
```diff
@@ -9,3 +9,67 @@ export function dateTimeFormat(timestamp, options) {
...
+  const filterViewTypes = searchParams.get('viewType');
+  const filterTypes = searchParams.get('type');
+  const filterNames = searchParams.get('name');
+  const filterKeyword = searchParams.get('keyword');
+  const filterTimestamp = searchParams.get('timeStamp');
+  const filterTabId = parseInt(searchParams.get('tabId'), 10) || null;
+
+  const ids = getJSONParseVal(filterIds);
+  let viewTypes = getJSONParseVal(filterViewTypes);
+  // converting empty string to undefined
```
Rob--W, line 55: This comment isn't accurate any more. Remove it.

src/lib/web-component/filter-option/filter-option-element.js
```diff
@@ -6,8 +6,8 @@ export class FilterOption extends HTMLElement {
     super();
 
     this.viewCheckboxLabels = new Set();
-    // It contains the checked checkboxes
-    this.activeCheckboxLabels = new Set();
+    // It contains the unchecked checkboxes
+    this.inactive = new Set();
```
Rob--W, line 10: This name is too vague. If you choose a meaningful name, the comment wouldn't be needed.
How about `uncheckedCheckboxLabels`?

src/lib/formatters.js
```diff
@@ -9,3 +9,67 @@ export function dateTimeFormat(timestamp, options) {
...
+    searchParams.set(key, searchParamVal);
+  });
+
+  return searchParams;
+}
+
+function getJSONParseVal(str) {
+  try {
+    return JSON.parse(str);
+  } catch (error) {
```
Rob--W, line 39: I suppose that you've created this to work around errors. Did you mean to include (new?) unit tests that trigger the error, or is that part of follow-up work?

atiqueahmedziad: Since you said to think about the edge cases (e.g. incorrect/missing input), I just put random search params in url and got an error from `JSON.parse`. This is the work arround for that. 
Should I add tests for this?

src/lib/web-component/filter-option/filter-option-element.js
atiqueahmedziad: Yeah, `uncheckedCheckboxLabels` sounds great. I also removed the comment.

src/lib/formatters.js
atiqueahmedziad: done :-)

src/lib/ext-activitylog.js
atiqueahmedziad: done :-)

src/lib/formatters.js
Rob--W: Yes. There are more ways to break the functionality (e.g. something that is valid JSON but not having the expected type), but you'll discover that in tests.

src/lib/ext-activitylog.js
```diff
@@ -183,6 +184,34 @@ class View {
     this.apiNameFilter.updateFilterCheckboxes(filteredLogs);
   }
 
+  setExistedFilters(updateFilter) {
```
Rob--W, line 187: I meant to comment about this before; You probably meant "set existing filters" instead of "set existed filters".

`setInitialFilter` may be a better name though, because it does seem that the method can be called only once (as the implementations are not yet prepared to deal with setting the filter after the initialization).

src/lib/formatters.js
```diff
@@ -9,3 +9,66 @@ export function dateTimeFormat(timestamp, options) {
...
+    return null;
+  }
+}
+
+export function deSerializeFilters(searchParams) {
+  const filterIds = searchParams.get('id');
+  const filterViewTypes = searchParams.get('viewType');
+  const filterTypes = searchParams.get('type');
+  const filterNames = searchParams.get('name');
+  const filterKeyword = searchParams.get('keyword');
```
Rob--W, line 49: `filterKeyword = ... || '';`

Do add a test. To find relevant edge cases, look at your code and try to see if the end result is guaranteed to have the expected type as defined in the Model at `filter`.

src/lib/web-component/filter-timestamp/filter-timestamp-element.js
```diff
@@ -53,6 +53,30 @@ export class FilterTimestamp extends HTMLElement {
...
+    }
+
+    if (timestamp?.stop != null) {
+      this.timeStamp.stop = timestamp.stop;
+      this.stopTimeLabel.textContent = dateTimeFormat(timestamp.stop);
+      this.filterContainer.hidden = false;
+      this.clearStopTimeBtn.hidden = false;
+    }
+
+    this.dispatchFilterChange();
```
Rob--W, line 77: Why is it necessary to call `dispatchFilterChange` here?

atiqueahmedziad: not needed. Updated :-)

src/lib/formatters.js
atiqueahmedziad: Updated and added test that guarantees the same type as defined in the model.

src/lib/ext-activitylog.js
atiqueahmedziad: Thanks. Updated the name as `setInitialFilter` :-)

tests/formatters.test.js
```diff
@@ -0,0 +1,78 @@
...
+    timeStamp,
+  };
+
+  const serializedFormat = formatters.serializeFilters(
+    searchParams,
+    updateFilter
+  );
+
+  expect(serializedFormat.get('id')).toEqual(JSON.stringify(extIds));
+  expect(serializedFormat.get('viewType')).toEqual(JSON.stringify(viewTypes));
```
Rob--W, line 31: Hardcode the string instead of using `JSON.stringify`. It's not obvious from the code to those unfamiliar with the exact working of `JSON.stringify` that the stringified value is `["popup",null]`.

tests/formatters.test.js
```diff
@@ -0,0 +1,78 @@
...
+
+  const searchParams = new URLSearchParams(testUrl.search.slice(1));
+  const filterObject = formatters.deSerializeFilters(searchParams);
+
+  expect(filterObject).toMatchObject(expectedFilterObject);
+});
+
+test('deSerializeFilters method should return the default filter object if there are no valid filters', () => {
+  const testUrl = new URL(
+    'http://test.html?invalidParam="wrongValue"&id=null&viewType="wrong-type"&timeStamp="wrong-type"&tabId=undefined'
```
Rob--W, line 62: The question is - does one invalid filter disable all other filters? You need to test a filter that includes at least one valid filter to check the difference.

And the test is not exhaustive, you need a test case with the filter value `{}` because it would show an issue in the implementation.

tests/formatters.test.js
atiqueahmedziad: Yeah, done :-)

tests/formatters.test.js
atiqueahmedziad: I have updated this test, now the test description is - `deSerializeFilters method should return the filter object with valid filter properties, ignoring the invalid filter properties`. Is this okay?

tests/formatters.test.js
```diff
@@ -0,0 +1,85 @@
...
+  const serializedFormat = formatters.serializeFilters(
+    searchParams,
+    updateFilter
+  );
+
+  const expectedExtIds = '["addon1@test.com","addon2@test.com"]';
+  const expectedViewTypes = '["popup",null]';
+  const expectedTypes = '["api_call","content_script"]';
+  const expectedNames = '["test@api"]';
+  const expectedTabId = '22';
```
Rob--W, line 34: This should have been `22`.

Use toStrictEqual instead of isEqual in your assertions to catch this kind of error.

tests/formatters.test.js
```diff
@@ -0,0 +1,85 @@
...
+    name: { exclude: new Set(['tabs.executeScript']) },
+    tabId: 982,
+    keyword: 'undefined',
+    timeStamp: { start: 123123, stop: 456456 },
+  };
+
+  const searchParams = new URLSearchParams(testUrl.search.slice(1));
+  const filterObject = formatters.deSerializeFilters(searchParams);
+
+  expect(filterObject).toMatchObject(expectedFilterObject);
```
Rob--W, line 63: I'm not sure, can you verify that .toMatchObject actually checks the contents of the Set? That is, if you intentionally make the expectation invalid, does the test fail as expected?

src/lib/formatters.js
```diff
@@ -9,3 +9,73 @@ export function dateTimeFormat(timestamp, options) {
...
+    return null;
+  }
+
+  try {
+    const parsedData = JSON.parse(str);
+    if (options?.returnArray) {
+      return Array.isArray(parsedData) ? parsedData : [];
+    } else if (options?.returnObject) {
+      return typeof parsedData === 'object' ? parsedData : null;
+    }
```
Rob--W, line 47: return value in other cases is missing.

Rob--W (approved):
Address the issues below and it's good to merge.

tests/formatters.test.js
atiqueahmedziad: I am getting the `"22"` since I am getting the tab id from search params. Should I parse that into number and then check ?

tests/formatters.test.js
atiqueahmedziad: Yes, it checks the content of the Set too. :-)

src/lib/formatters.js
atiqueahmedziad: added now :-)

tests/formatters.test.js
Rob--W: I don't think that you're getting `"22"` but 22. You're parsing `tabId` before returning it:

```javascript
const filterTabId = parseInt(searchParams.get('tabId'), 10) || null;
```

Rob--W (approved):

tests/formatters.test.js
atiqueahmedziad: > I don't think that you're getting `"22"` but 22. You're parsing `tabId` before returning it:
> 
> ```js
> const filterTabId = parseInt(searchParams.get('tabId'), 10) || null;
> ```

Yes, I have added this parsing now.

PR #51, opened 2020-08-31
test: Improved test coverage

### LINKED ISSUE #50: Increase test coverage
Lots of files have complete coverage! 🎉 

Let's add some test coverage for the partly covered files: https://coveralls.io/github/mozilla/extension-activity-monitor

atiqueahmedziad:
Fixes #50

- [x] Turn the code coverage of `ext-listen.js` to 100%
- [x] Turn the code coverage of `ext-monitor.js` to 100%
- [x] Turn the code coverage of `ext-activitylog.js` to 100%

### DIFF
--- a/src/lib/ext-activitylog.js
+++ b/src/lib/ext-activitylog.js
@@ -156,28 +156,32 @@ class View {
   }
 
   handleEvent(event) {
-    if (event.type === 'click') {
-      switch (event.target) {
-        case this.clearLogBtn:
-          this.clearLogBtn.dispatchEvent(new CustomEvent('clearlog'));
-          break;
-        case this.optionsBtn:
-          dropDownController.toggleDropDown(this.optionsBtn);
-          break;
-        case this.saveLogBtn:
-          this.saveLogBtn.dispatchEvent(new CustomEvent('savelog'));
-          break;
+    try {
+      if (event.type === 'click') {
+        switch (event.target) {
+          case this.clearLogBtn:
+            this.clearLogBtn.dispatchEvent(new CustomEvent('clearlog'));
+            break;
+          case this.optionsBtn:
+            dropDownController.toggleDropDown(this.optionsBtn);
+            break;
+          case this.saveLogBtn:
+            this.saveLogBtn.dispatchEvent(new CustomEvent('savelog'));
+            break;
+        }
+      } else if (event.type === 'change' && event.target === this.loadLogFile) {
+        const logFile = event.target.files[0];
+        this.loadLogFile.value = '';
+        this.loadLogFile.dispatchEvent(
+          new CustomEvent('loadlog', { detail: logFile })
+        );
+      } else if (event.type === 'logcountchange') {
+        this.updateLogCounter(event.detail);
+      } else {
+        throw new Error(`wrong event type - ${event.type}`);
       }
-    } else if (event.type === 'change' && event.target === this.loadLogFile) {
-      const logFile = event.target.files[0];
-      this.loadLogFile.value = '';
-      this.loadLogFile.dispatchEvent(
-        new CustomEvent('loadlog', { detail: logFile })
-      );
-    } else if (event.type === 'logcountchange') {
-      this.updateLogCounter(event.detail);
-    } else {
-      throw new Error(`wrong event type - ${event.type}`);
+    } catch (error) {
+      this.setError(error.message);
     }
   }
 
@@ -299,19 +303,7 @@ class Controller {
       this.view.saveLogBtn.addEventListener('savelog', this);
 
       browser.runtime.connect({ name: 'monitor-realtime-logs' });
-      browser.runtime.onMessage.addListener((message) => {
-        const { requestTo, requestType } = message;
-
-        if (requestTo !== 'activity-log') {
-          return;
-        }
-
-        if (requestType === 'appendLogs') {
-          this.handleNewLogs([message.log]);
-        } else {
-          throw new Error(`wrong request type found - ${requestType}`);
-        }
-      });
+      browser.runtime.onMessage.addListener(this.onMessageListener);
 
       if (document.location.search) {
         const updateFilter = deSerializeFilters(searchParams);
@@ -327,6 +319,22 @@ class Controller {
     }
   }
 
+  onMessageListener = (message) => {
+    const { requestTo, requestType } = message;
+
+    if (requestTo !== 'activity-log') {
+      return;
+    }
+
+    if (requestType === 'appendLogs') {
+      this.handleNewLogs([message.log]);
+    } else {
+      return Promise.reject(
+        new Error(`wrong request type found - ${requestType}`)
+      );
+    }
+  };
+
   async getExistingLogs() {
     const { existingLogs } = await browser.runtime.sendMessage({
       requestType: 'sendAllLogs',
@@ -341,21 +349,25 @@ class Controller {
   }
 
   handleEvent(event) {
-    switch (event.type) {
-      case 'loadlog':
-        this.loadLogs(event.detail);
-        break;
-      case 'savelog':
-        this.saveLogs();
-        break;
-      case 'filterchange':
-        this.onFilterChange(event.detail);
-        break;
-      case 'clearlog':
-        this.handleClearLogs();
-        break;
-      default:
-        throw new Error(`wrong event type found - ${event.type}`);
+    try {
+      switch (event.type) {
+        case 'loadlog':
+          this.loadLogs(event.detail);
+          break;
+        case 'savelog':
+          this.saveLogs();
+          break;
+        case 'filterchange':
+          this.onFilterChange(event.detail);
+          break;
+        case 'clearlog':
+          this.handleClearLogs();
+          break;
+        default:
+          throw new Error(`wrong event type found - ${event.type}`);
+      }
+    } catch (error) {
+      this.view.setError(error.message);
     }
   }
 
--- a/src/lib/ext-activitylog.js
+++ b/src/lib/ext-activitylog.js
@@ -1,5 +1,4 @@
 import dropDownController from './DropDownController.js';
-import { save } from './save-load.js';
 import { serializeFilters, deSerializeFilters } from './formatters.js';
 
 class Model {
@@ -156,32 +155,28 @@ class View {
   }
 
   handleEvent(event) {
-    try {
-      if (event.type === 'click') {
-        switch (event.target) {
-          case this.clearLogBtn:
-            this.clearLogBtn.dispatchEvent(new CustomEvent('clearlog'));
-            break;
-          case this.optionsBtn:
-            dropDownController.toggleDropDown(this.optionsBtn);
-            break;
-          case this.saveLogBtn:
-            this.saveLogBtn.dispatchEvent(new CustomEvent('savelog'));
-            break;
-        }
-      } else if (event.type === 'change' && event.target === this.loadLogFile) {
-        const logFile = event.target.files[0];
-        this.loadLogFile.value = '';
-        this.loadLogFile.dispatchEvent(
-          new CustomEvent('loadlog', { detail: logFile })
-        );
-      } else if (event.type === 'logcountchange') {
-        this.updateLogCounter(event.detail);
-      } else {
-        throw new Error(`wrong event type - ${event.type}`);
+    if (event.type === 'click') {
+      switch (event.target) {
+        case this.clearLogBtn:
+          this.clearLogBtn.dispatchEvent(new CustomEvent('clearlog'));
+          break;
+        case this.optionsBtn:
+          dropDownController.toggleDropDown(this.optionsBtn);
+          break;
+        case this.saveLogBtn:
+          this.saveLogBtn.dispatchEvent(new CustomEvent('savelog'));
+          break;
       }
-    } catch (error) {
-      this.setError(error.message);
+    } else if (event.type === 'change' && event.target === this.loadLogFile) {
+      const logFile = event.target.files[0];
+      this.loadLogFile.value = '';
+      this.loadLogFile.dispatchEvent(
+        new CustomEvent('loadlog', { detail: logFile })
+      );
+    } else if (event.type === 'logcountchange') {
+      this.updateLogCounter(event.detail);
+    } else {
+      this.setError(`wrong event type - ${event.type}`);
     }
   }
 
@@ -283,17 +278,11 @@ class Controller {
       this.view.contentWrapper.classList.add('load-logs');
 
       const currentTab = await browser.tabs.getCurrent();
-      let logs;
 
-      try {
-        logs = await browser.runtime.sendMessage({
-          requestType: 'getLoadedLogs',
-          requestTo: 'ext-monitor',
-          requestParams: { tabId: currentTab.id },
-        });
-      } catch (error) {
-        this.view.setError(error.message);
-      }
+      const logs = await this.sendMsgToBackground({
+        requestType: 'getLoadedLogs',
+        requestParams: { tabId: currentTab.id },
+      });
 
       if (logs?.length) {
         this.handleNewLogs(logs);
@@ -336,10 +325,10 @@ class Controller {
   };
 
   async getExistingLogs() {
-    const { existingLogs } = await browser.runtime.sendMessage({
+    const { existingLogs } = await this.sendMsgToBackground({
       requestType: 'sendAllLogs',
-      requestTo: 'ext-monitor',
     });
+
     return existingLogs;
   }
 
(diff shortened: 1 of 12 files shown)


wagnerand:
@atiqueahmedziad Hey, just checking in on whether you are planning to finish this PR? Thanks!

atiqueahmedziad:
> @atiqueahmedziad Hey, just checking in on whether you are planning to finish this PR? Thanks!

Yes, sorry for the silence. I will be updating the PR shortly.

atiqueahmedziad:
Just to note, I am still working on it. I will comment here when I am finished with adding the tests.

src/lib/ext-activitylog.js
```diff
@@ -156,28 +156,32 @@ class View {
...
-      this.loadLogFile.value = '';
-      this.loadLogFile.dispatchEvent(
-        new CustomEvent('loadlog', { detail: logFile })
-      );
-    } else if (event.type === 'logcountchange') {
-      this.updateLogCounter(event.detail);
-    } else {
-      throw new Error(`wrong event type - ${event.type}`);
+    } catch (error) {
+      this.setError(error.message);
```
rpl, line 184: would you mind to describe me the reasons why we are wrapping all the handleEvent method (this one and also the one defined at line 351) in a try...catch? 
is it a legit scenario or just something happening in the test case? 

(I'm basically wondering if it is the test case that should provide a more realistic testing environment, otherwise we be handling errors at every "layer", https://en.wikipedia.org/wiki/Defensive_programming may be an interesting read on the topic)

rpl (changes requested):
Hi @atiqueahmedziad 
I was so happy to see you coming back to this PR for some more updates, based on a quick look to the last version of this PR it seems to be still pending some more changes (e.g. there are some empty or commented out tests and some inline todo comments), is that right?.

Anyway, I did notice that I had one feedback comment here that I never submitted (eh, my apologies) and I collected a couple more (some questions and some side notes that I think it may be useful for you to keep in mind).

I'll need to refresh my memory about this PR and the in general the status of the project to give this a more complete review round, I'll try to do that upfront If I can reach it, but let me know when this is ready for a proper review round by renew your review request.

Also, feel free to ask me any questions and/or doubts you may already have (about the changes you were already planning or about the questions and side notes I submitted as part of this round of feedback)

atiqueahmedziad:
I am very sorry for address this after so long. (I was having some tough time in the middle).
Thanks a lot for the comments @rpl . However, I have not addressed the feedbacks yet. I have just finished adding tests but need to do some cleanup to make it suitable for review. I will add another comment/ping you in the matrix within the next 2-3 days (when I am finished).

src/lib/ext-activitylog.js
atiqueahmedziad: I just rechecked that part of the code. There should no error thrown in the try catch block at line 183. So, I have removed this try and catch block.

At line 351, there is no need of try catch block since I have been handling the errors inside every methods (which have been called from `handleEvent`) separately. I made a helper function to catch the error from `sendMessage` API at a single place. The helper function communicates with the backend.
```javascript
// helper function
async sendMsgToBackground({ requestType, requestParams }) {
  try {
    return await browser.runtime.sendMessage({
      requestTo: 'ext-monitor',
      requestType,
      requestParams,
    });
  } catch (error) {
    this.view.setError(error.message);
  }
}

```
You can view the changes in this commit - 6e0583203

tests/ext-listen.test.js
```diff
@@ -58,6 +61,23 @@ test('openActivityLogPage should create tab with ActivityLogPageURL', async () =
...
+});
+
+test('ActivityLogPageURL method should return url with search params if search params are available', () => {
+  const getURL = jest.fn();
+  const create = jest.fn();
+
+  window.browser = {
+    runtime: { getURL },
+    tabs: { create },
+  };
```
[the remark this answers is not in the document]
atiqueahmedziad: Thanks for noticing. I have removed it and other tests which were unnecessary here - 9d8c5a68f

src/lib/web-component/filter-timestamp/filter-timestamp-element.js
```diff
@@ -25,9 +25,7 @@ export class FilterTimestamp extends HTMLElement {
     this.startTimeLabel = this.dropDownList.querySelector('#startTimeLabel');
     this.stopTimeLabel = this.dropDownList.querySelector('#stopTimeLabel');
 
-    this.clearFilterBtn = filterWrapper.querySelector('#clearFilter');
```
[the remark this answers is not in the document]
atiqueahmedziad: Sorry for not commenting on this change earlier. But thanks for asking. 

Previously (before we had introduced the new UI for EAM), we had 3 clear options for timestamp filters, such as- 
1. Clearing the Start Time filter
2. Clearing the End Time filter
3. Clearing both timestamp filters.

This `clearFilterBtn` was for clearing the both timestamp filters. With the introduction of the new UI of EAM, we have only first two options. But I had mistakenly forget to remove `clearFilterBtn` it then. We have no use of `clearFilterBtn` currently. So, I have removed it while cleaning the codes.

### Current UI of timestamp filter
Screenshot 2021-04-05 at 8 41 21 PM

tests/ext-listen.test.js
```diff
@@ -18,6 +18,7 @@ test('getMonitorStatus should have requestType "getMonitorStatus" and resolves w
 
   expect(sendMessage.mock.calls[0][0]).toMatchObject({
     requestType: 'getMonitorStatus',
+    requestTo: 'ext-monitor',
```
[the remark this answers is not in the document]
atiqueahmedziad: Yes, indeed this helping function is very helpful. I have used this helping function in multiple files.

> Also, try to not drive the new tests creation just based on code coverage:
> 
>     * personally I usually plan the set of tests based on the scenarios that should be tested, and just use the coverage reports to evaluate what scenario I missed in the initial plan.
> 
>     * then, once I'm happy about the set of scenarios that I've already covered with the test cases, I look to the coverage reports again to see what is still missing (and I evaluate if it is worth adding a new separate test case for a new particular scenario or if the coverage could just be part of an additional step in an existing test).
> 
>     * If some of the new test scenarios I'm introducing looks too similar to other test cases, I refactor out the shared parts in a new test helper (and parametrize the parts that needs to be different) and then I rewrite all the similar tests by reusing the new test helper

Thanks a lot for these opinions. After reading this part of your comment, I went through every test files, rechecked them and re-organized them based on the scenarios that I see while operating the extension. After that I wrote a few tests for the sake of coverage, for example: throwing errors intentionally to check if it has been caught by some lines, which were untested before. I also removed somes tests here - 9d8c5a68f, which were unnecessary. I think the tests are in a better state now. Thanks a lot @rpl

src/lib/web-component/filter-timestamp/filter-timestamp-element.js
```diff
@@ -163,8 +158,8 @@ export class FilterTimestamp extends HTMLElement {
   }
 
   disconnectedCallback() {
-    browser.menus.onClicked.removeListener(this.setFilterRange);
-    browser.menus.onHidden.removeListener(this.onHiddenListener);
+    browser.menus?.onClicked.removeListener(this.setFilterRange);
+    browser.menus?.onHidden.removeListener(this.onHiddenListener);
```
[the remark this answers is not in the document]
atiqueahmedziad: Sorry, I haven't noticed it carefully. Thank you for noticing. We shouldn't unsubscribe any of the listeners when we aren't subscribe to any.

In `connectedCallback()` we are subscribing all the listeners only when the `menus` API is available. I think we should do the same in case of unsubscribing the listeners in `disconnectedCallback()` i.e. we shouldn't unsubscribe to any listeners if  `menus` API is missing. I have done this in d3fa9e1.

src/lib/web-component/filter-timestamp/filter-timestamp-element.js
```diff
@@ -132,9 +130,6 @@ export class FilterTimestamp extends HTMLElement {
   handleEvent(event) {
     if (event.type === 'click') {
       switch (event.target) {
-        case this.clearFilterBtn:
-          this.onClearFilter();
-          break;
```
atiqueahmedziad, line 137: Removing this since in the current UI we aren't removing both the timestamp at once. Explanation given here - https://github.com/mozilla/extension-activity-monitor/pull/51/files#r607192050

rpl:
> Thanks a lot for these opinions. After reading this part of your comment, I went through every test files, rechecked them and re-organized them based on the scenarios that I see while operating the extension. After that I wrote a few tests for the sake of coverage, for example: throwing errors intentionally to check if it has been caught by some lines, which were untested before. I also removed somes tests here - 9d8c5a6, which were unnecessary. I think the tests are in a better state now. Thanks a lot @rpl

@atiqueahmedziad Apologies for not being able to give this another review pass yet, I'll try to get to it asap, but in the meantime I wanted to say you that: I have been really happy to see you coming back to this PR for a round of new changes, and I'm really glad to hear that my comment describing how I usually approach tests was useful to let you rethink and reorganize the test cases around scenarios instead of just code coverage.

wagnerand:
@rpl friendly ping. Anything left to do here?

rpl:
> @rpl friendly ping. Anything left to do here?

yes, I have to give this a proper review pass, I haven't forgot about this (I do actually have it explicitly listed in my local backlog list of pending review) but I wasn't able to reach it yet.

I'll try to get back to this review as soon as possible.

src/lib/web-component/filter-option/filter-option-element.js
```diff
@@ -95,8 +95,6 @@ export class FilterOption extends HTMLElement {
       }
 
       this.dispatchFilterChangeEvent();
-    } else {
-      throw new Error(`wrong event type - ${event.type}`);
     }
```
rpl, line 98: @atiqueahmedziad what are the reasons behind this removal? was this actually triggered in any of the test cases?

src/lib/web-component/filter-timestamp/filter-timestamp-element.js
```diff
@@ -163,6 +158,12 @@ export class FilterTimestamp extends HTMLElement {
   }
 
   disconnectedCallback() {
+    // context menu doesn't work on devtool_page
+    // see: https://github.com/mozilla/extension-activity-monitor/issues/43
+    if (!browser.menus?.onClicked || !browser.menus?.onHidden) {
+      return;
+    }
+
```
rpl, line 166: @atiqueahmedziad it may be better to do the opposite here and change the two calls to `browser.menus` to make sure using the optional chaining operator so that they will become a no-op if `browser.menus` isn't available. I think this is less risky because by returning earlier we may mistakenly skip also some other call that should instead be called (e.g. in this particular case we may still want to call `this.filterToggleBar.removeEvnetListener` and `this.dropDownList.removeEventListener`).

It should look something like:
```
    browser.menus?.onClicked.removeListener(this.setFilterRange);
    browser.menus?.onHidden.removeListener(this.onHiddenListener);
```

tests/ext-activitylog.test.js
```diff
@@ -70,113 +757,123 @@ test('show/hide logs associated with extension id that is checked/unchecked from
...
-  ).toBeTruthy();
+  const { logDetailWrapper } = activityLog.view.logView;
+  // Initially the log details is hidden (when no table is clicked)
+  expect(logDetailWrapper.hidden).toBe(true);
+  // clicking a row in the table to show logs details
+  row.click();
+  expect(logDetailWrapper.hidden).toBe(false);
+
+  clearLogBtn.click();
+  await observeChange(logDetailWrapper);
```
rpl, line 792: I see that elsewhere observeChange was called before triggering the change to the DOM elements and its promise returned value stored to be awaited on after triggering the side effect (e.g. lines 771-773 in this same file diff), is there a risk for this one to become an intermittent? was the one 771-773 failing intermittently if we would do the same there?

rpl (changes requested):
@atiqueahmedziad I'm sorry that this is taking me so long, but I do have to refresh my memory about what was the status of this PR and of the project in general.

I do have collected some more comments (including some changes requested on the implementation side and some question to clarify some parts of this PR that I wasn't completely sure where they fit).

The tests looks a bit repetitive, I think there may still be some more margin for further simplifying them.

e.g. code for setting up the test environment that may be taking quite some lines and repeating more then a couple of times, (which we may still move into test helpers shared by more test files or more test cases). 

To provide some more precise pointers I have to "read and get a full picture" of the new tests a bit more, and also pull the changes in this PR in my local clone and then give them a look and a try locally (in particular to make it easier for me to look the tests fragments side by side, and to refresh my memory about the current UI and what the new tests are meant to cover and then how they are currently organized).

I plan to come back to this again in the next few days, in the meantime feel free to reply to my comments below (and/or handle the changes requested).

You may also add some github comments around code in this PR that you think we may extract into test helpers, and mention me in the comment, this way I may come to it sooner and help you to weight the options we have in terms of refactoring that code into some more shared test helpers (and evaluate together if and how that would help us to simplify the test cases).

src/lib/web-component/filter-option/filter-option-element.js
atiqueahmedziad: No, this hasn't been triggered by any of the test cases.
Currently, we are not catching this error anywhere (we also didn't set any UI to show this error in the `filter-option` web component) and this error is very less likely to get produced.

src/lib/web-component/filter-timestamp/filter-timestamp-element.js
atiqueahmedziad: I agree. :-) Thanks for the suggestion

tests/ext-activitylog.test.js
atiqueahmedziad: @rpl Nice catch! It felt a bit weird to see this code at line no. 792. I think the code at line no. 792 should follow the same way of resolving promises once the DOM changes like the one at line 771-773 i.e.
- Setup Promise with current state of DOM
- Trigger DOM changes
- Wait for the promise to resolve

Currently, it's not failing because when the `clearLogBtn.click()` event occurs, it does some work i.e. remove logs from background (includes some promise resolving), then remove table rows from DOM. It may fail intermittently if the `clearLogBtn.click()` event had to do very less work. 

> was the one 771-773 failing intermittently if we would do the same there?

No, it's not failing since the `ActivityLog()` have to do some work i.e. fetch all the logs from background (includes some promise resolving) before making changes in the DOM.

atiqueahmedziad:
@rpl I will be adding some comments in the PR regarding the test helpers the tests by tomorrow.

tests/ext-activitylog.test.js
```diff
@@ -27,28 +27,715 @@ function observeChange(element) {
...
+        },
+        {
+          /* renders the 2nd row in the table */
+          id: 'id2@test',
+          viewType: 'viewType2@test',
+          type: 'type2@test',
+          data: { test: 'test2@data' },
+          timeStamp: 1597686226302,
+        },
+      ];
```
atiqueahmedziad, line 61: @rpl I want to take this `logs` to the global scope (may need to 1 or 2 more log objects in the array) and reuse whenever needed.

tests/ext-activitylog.test.js
```diff
@@ -27,28 +27,715 @@ function observeChange(element) {
...
+        runtime: {
+          onMessage: { addListener },
+          sendMessage,
+          connect,
+        },
+      };
+
+      sendMessage.mockImplementation(() => {
+        return Promise.resolve({ existingLogs: [] });
+      });
```
atiqueahmedziad, line 77: @rpl I want to have a common the `browser` object that serves all the test cases and and may be set it with `beforeAll` to run for a  single time before running all the test cases.

tests/ext-activitylog.test.js
```diff
@@ -27,28 +27,715 @@ function observeChange(element) {
...
+
+      document.body.innerHTML = activityLogBody;
```
atiqueahmedziad, line 79: This line maybe suitable to execute before each of the test cases i.e. with `beforeEach`

atiqueahmedziad (commented):
@rpl On a quick check, I find these redundant code in every test cases which we may extract these into test helpers to simplify the test cases. Let me know what you think about these.

tests/ext-activitylog.test.js
```diff
@@ -27,28 +27,715 @@ function observeChange(element) {
...
+    document.body.innerHTML = activityLogBody;
+
+    const { activityLog } = new ActivityLog();
+
+    activityLog.handleNewLogs(logs);
+
+    const tableBody = activityLog.view.logView.shadowRoot.querySelector(
+      'tbody'
+    );
+    const getTableRows = () => tableBody.querySelectorAll('tr');
```
atiqueahmedziad, line 310: @rpl This lines i.e.
```
const tableBody = activityLog.view.logView.shadowRoot.querySelector('tbody');
const getTableRows = () => tableBody.querySelectorAll('tr');
```
 is redundant in in many test cases including every tests under `filtering logs from the logs table` section. Maybe it's better to extract it as test helper.

wagnerand:
Hey @atiqueahmedziad, we are closing this pull request for inactivity, feel free to reopen when you have worked on it again.

Issue #32, opened 2020-08-07
Log counter should be displayed


atiqueahmedziad:
There should be a log counter which will display -
- The total number of logs are collected.
- The number of logs that did match the filter.

The counter should be displayed nearby the log view.

PR #52, opened 2020-09-09
feat: new layout implemented and existing styles improved

### LINKED ISSUE #21: Improvements to style and layout applied to the activity log page
As mention in #15 the activitylog page, once the activitylog page is complete enough from a features perspective, we could apply to the page some style and layouting improvements (e.g. add a proper page layout based on css grid layout, make the table header fixed while scrolling the table content etc).

atiqueahmedziad:
fixes #21 

#### Filter options
Screenshot 2020-09-09 at 9 46 00 PM

#### "Gear button" popup (top right side)
Screenshot 2020-09-09 at 9 48 00 PM

### Loaded Logs
Screenshot 2020-09-09 at 9 48 46 PM

### DevTools page

Screenshot 2020-09-09 at 9 40 06 PM


Rob--W (changes requested):
- Why is the "dropdown" menu not closing when something else is clicked / focused?
- When a dropdown doesn't contain any content, it would be nice to render one dummy value instead.
- In small windows (e.g. resized to be 1400px wide), the timestamp box (and other boxes) is too narrow. The rendered time spans 4 lines. Is it possible to make the box wider (maybe in a follow-up?)? Take a look at the `<select>` element for inspiration.
- It is not visually obvious that a row is selected.
- The top bar, the heading and the content are too similar, which makes it a bit difficult to separate the content area from the rest of the UI.
- The bottom bar looks odd - what's the purpose of `Collected Logs: |`? From looking at the code it seems that you meant to add more active functionality.

atiqueahmedziad:
> * The top bar, the heading and the content are too similar, which makes it a bit difficult to separate the content area from the rest of the UI.
Can you please explain above point a bit? Is it about the white color (which has been used as background of table rows and filter options) ?

> * The bottom bar looks odd - what's the purpose of `Collected Logs: |`? From looking at the code it seems that you meant to add more active functionality.
Currently in code, I haven't put any dummy value. It could be like `Collected Logs: 1122 | 150` where 1122 is number of total logs and 150 is number of filtered logs. 
I just designed the UI. It will be displayed correctly once I fix the #32 . I will add a PR fixing #32 very soon.

Rob--W:
> > * The top bar, the heading and the content are too similar, which makes it a bit difficult to separate the content area from the rest of the UI.
> 
> Can you please explain above point a bit? Is it about the white color (which has been used as background of table rows and filter options) ?

Sort of, yes. The top bar does not visually stand out as it looks very similar to a row in the table.

atiqueahmedziad:
> * Why is the "dropdown" menu not closing when something else is clicked / focused?

Now the filter dropdowns are closed when something else is clicked.

> * When a dropdown doesn't contain any content, it would be nice to render one dummy value instead.

"No Filter options" label is being added when no filters options are available. Also, when there are no logs in the table, "No Logs Found" label is been added.

> * In small windows (e.g. resized to be 1400px wide), the timestamp box (and other boxes) is too narrow. The rendered time spans 4 lines. Is it possible to make the box wider (maybe in a follow-up?)? Take a look at the `<select>` element for inspiration.

Yeah, doing it in a follow-up.

> * It is not visually obvious that a row is selected.

Now when a row is selected, it gets highlighted.

> * The top bar, the heading and the content are too similar, which makes it a bit difficult to separate the content area from the rest of the UI.

I tried with adding different colors to the dropdown and table rows, but that didn't look good. So, I have restyled the `box-shadow` property of the dropdown which seems fine (to able to separate the different areas)  to my eye. Will you kindly check if you can differentiate the layouts?

Rob--W (commented):
- Bug: The even-odd coloring of rows doesn't work when filters are toggled: visually adjacent rows have the same background color.
- The way that you're hiding dropdown boxs is quite complicated and has a bug (or intentional feature?); when a different dropdown (or some buttons such as the clear button) is clicked, the previous dropdown does not close. The management of dropdowns is not critical functionality that belongs to the View class. To improve it, you could consider taking advantage of the fact that only one dropdown is supposed to be visible at any time.

src/lib/web-component/log-view/log-view.css
```diff
@@ -78,3 +73,22 @@ table td {
...
+  width: 100%;
+  height: calc(100% - 120px);
+  display: flex;
+  align-items: center;
+  justify-content: center;
+  font-size: 25px;
+  font-weight: 700;
+}
+
+/* hidden attribute is not being applied to the elements inside the shadow DOM */
```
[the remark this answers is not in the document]
atiqueahmedziad: I have updated this comment.

src/activitylog/activitylog.css
```diff
@@ -1,36 +1,179 @@
...
+.options-dropdown img {
+  width: 24px;
+  height: auto;
+  padding-right: 8px;
 }
 
-.failure {
-  background-color: #961616;
-  padding: 10px;
+.options-dropdown:after,
```
Rob--W, line 118: Use `::after` and `::before` instead of `:after` and `:before`.

They are equivalent and will be supported because of back-compat, but a singular `:` is now used for pseudo-classes, whereas a double `::` is used for pseudo-elements.

For more context, see https://stackoverflow.com/questions/7327326/what-is-the-difference-between-before-and-before

atiqueahmedziad:
>     * Bug: The even-odd coloring of rows doesn't work when filters are toggled: visually adjacent rows have the same background color.

This is not done yet. I will have to look at how to fix this (might need some suggestions). 

>     * The way that you're hiding dropdown boxs is quite complicated and has a bug (or intentional feature?); when a different dropdown (or some buttons such as the clear button) is clicked, the previous dropdown does not close. The management of dropdowns is not critical functionality that belongs to the View class. To improve it, you could consider taking advantage of the fact that only one dropdown is supposed to be visible at any time.

Yeah, this was an intentional feature to keep the other filters opened when the user clicks on different filter options but it's a bit complicated. So, currently, only one dropdown can be visible at any time.

src/lib/ext-activitylog.js
```diff
@@ -226,8 +257,19 @@ class View {
...
+    this.pageType.textContent = heading;
+  }
+
+  updateMenus({ pageType }) {
+    if (pageType === 'load-logs') {
+      this.clearLogBtn.hidden = true;
+      this.optionsBtn.hidden = true;
+      this.menuWrapper
+        .querySelectorAll('.separator')
+        .forEach((eachSeparator) => (eachSeparator.hidden = true));
```
Rob--W, line 270: I don't understand why the separator needs to be hidden unconditionally in this case. Why would the source of the log dictate whether separators are shown or not?

src/lib/ext-activitylog.js
```diff
@@ -167,6 +182,21 @@ class View {
...
+    if (eventTarget !== this.optionsBtn) {
+      this.optionsDropdown.hidden = true;
+    }
+
+    if (filterOptionsSet.has(eventTarget)) {
+      filterOptionsSet.delete(eventTarget);
+    }
+
+    for (const filterOption of filterOptionsSet) {
+      filterOption?.toggleDropdown({ displayPref: 'hide' });
```
Rob--W, line 196: Why `?.` ? Does the set ever contain void/null values?

src/lib/ext-activitylog.js
```diff
@@ -147,15 +159,18 @@ class View {
...
-          throw new Error(`wrong event target - ${event.target.tagName}`);
+        case this.optionsBtn:
+          this.optionsDropdown.hidden = !this.optionsDropdown.hidden;
+          break;
+        case this.saveLogBtn:
+          this.saveLogBtn.dispatchEvent(new CustomEvent('savelog'));
+          break;
       }
+
+      this.hideDropdowns(event.target);
```
Rob--W, line 173: The `hideDropdowns` logic works but it requires a global click listener runs too often. In https://github.com/mozilla/extension-activity-monitor/pull/52#pullrequestreview-486879126 I said "consider taking advantage of the fact that only one dropdown is supposed to be visible at any time."

I was thinking of this:

```
class DropDownController {
  showDropDown(elem) {
    this.hideDropDown();
    this.elem = elem;
    document.addEventListener("click", this);
    // TODO: Actually show elem
  }
  hideDropDown() {
    let {elem} = this;
    if (!elem) {
      return;
    }
    this.elem = null;
    document.removeEventListener("click", this);
    // TODO: Actually hide elem
  }
}
```

This is much simpler than your current implementation where you need to keep track of all elements.

Rob--W (changes requested):
The UI looks much more polished in this version, nice.

Here is some more feedback.

atiqueahmedziad:
The changes in commit ee0c936  are:
1. Hided separator and other elements with css instead of js
2. The colors are updated taking preference from [photon style guide](https://design.firefox.com/photon/visuals/color.html)
3. Aligned "Filtered by tab id" on the left in the bottom bar.
4. Made the content script url fit in API Name column. The viewtype column is empty for `content_script`.
5. Added vertical line in table view and text are aligned to the left.
6. Added `DropdownController` class in `ext-activitylog.js` which control the opening and closing of the dropdowns.
7. Load logs functionality is made available in activity logs page of "log loaded from file" mode.
8. Found which property is overriding the `hidden` attribute. It's the `flex` property. Updated the comments accordingly.
9. Filter options spacing updated. Now Timestamp filter option has more space then other filter options (as it requires so).

src/lib/ext-activitylog.js
atiqueahmedziad: Thanks for this. I have implemented like this.

src/activitylog/activitylog.css
atiqueahmedziad: Yes used `::` instead of single `:`

src/activitylog/activitylog.html
```diff
@@ -24,41 +24,68 @@
...
+        </div>
+      </div>
+      <div class="notice"></div>
+      <div class="log-view-wrapper">
+        <log-view />
+      </div>
+      <div class="footer-wrapper">
+        <div class="logs-counter">
+          <span title="Visible Logs">0</span>
+          <div class="separator"></div>
```
Rob--W, line 80: This should be a slanted separator `/`.

The common UI pattern is `x / y`, meaning `x out of y`.

src/lib/ext-activitylog.js
```diff
@@ -135,9 +175,31 @@ class View {
...
+    this.viewTypeFilter.addEventListener('triggerdropdown', this);
+    this.apiTypeFilter.addEventListener('triggerdropdown', this);
+    this.apiNameFilter.addEventListener('triggerdropdown', this);
+    this.timestampFilter.addEventListener('triggerdropdown', this);
+    this.optionsBtn.addEventListener('triggerdropdown', this);
+
+    this.optionsBtn.toggleDropdown = ({ displayPref }) => {
+      switch (displayPref) {
+        case 'show':
+          this.optionsDropdown.hidden = false;
```
Rob--W, line 194: The `optionsDropdown` element is a child of the `optionsBtn` element. There is no need for a custom JS method to toggle the visibility, please use CSS. This applies to almost every part in this patch that toggles the `hidden` attribute.

src/lib/ext-activitylog.js
```diff
@@ -135,9 +175,31 @@ class View {
     this.keywordFilter = document.querySelector('filter-keyword');
     this.timestampFilter = document.querySelector('filter-timestamp');
 
+    this.optionsBtn.addEventListener('click', this);
     this.clearLogBtn.addEventListener('click', this);
     this.saveLogBtn.addEventListener('click', this);
     this.loadLogFile.addEventListener('change', this);
+    this.logView.addEventListener('logcountchange', this);
+
+    this.extFilter.addEventListener('triggerdropdown', this);
```
Rob--W, line 184: Every custom element with "dropdown" functionality should manage the dropdown logic itself. This logic is an implementation detail of the custom component, and therefore you should not put listeners in the view here. The components themselves should invoke the DropDownController where needed.

The DropDownController is a singleton; since the components need a reference to an instance, the DropDownController can either export an instance, or the singleton design pattern can be used (`getInstance`).

src/lib/web-component/filter-option/filter-option-element.js
```diff
@@ -49,6 +51,8 @@ export class FilterOption extends HTMLElement {
   }
 
   addNewCheckbox(checkboxLabel) {
+    this.emptyFilterLabel.hidden = true;
```
Rob--W, line 54: This can be done with pure CSS:

- Either with the element, by letting the element be the last child and using the `+` selector to target the last label if it's hidden.
- Or with the `:empty` selector and pseudo-elements.

... or by just removing the filter content at the first run.

Currently, if the filter doesn't match anything, the table is still empty. Is that intentional?

Rob--W (changes requested):

src/lib/web-component/filter-option/filter-option.css
```diff
@@ -2,63 +2,69 @@
...
 .filter-checkbox input:checked:focus ~ .select-indicator {
-  background: #0e647d;
-}
-.filter-checkbox input:disabled ~ .select-indicator {
-  background: #e6e6e6;
-  opacity: 0.6;
-  pointer-events: none;
+  background: #008ea4;
 }
+
```
rpl, line 47: @atiqueahmedziad To make it easier to change the tweak the colors (at the moment to better match the color being used in the devtools panels, but also to make it possible to easily switch between themes, e.g. light and dark themes) it would be better to use CSS variables for the photon colors (which would also make it way easier to recognize which photon color a certain element is using) and also some CSS variables to separate the "kind of usage that a certain color is used for" from the actual photon color we are going to currently use for that.

I inspected the devtools panels (using the Browser Toolbox as I did suggest in our last meeting and also by looking for those variables on searchfox.org) and looked for the colors being used in the devtools panels and then I tried to quickly add some CSS variables on top of the changes in this PR without changing too much or the rest of the CSS rules.

Here is the comparison screenshot of the resulting activitylog page and the devtools panel:

image

There are some more tweaks that may be reasonable to apply to match the devtools panel look and feel even more (where/when it makes sense), but this looks already as a good start from my perspective.

It is ok if you prefer to merge this pull request as is (after handling the last round of review comments from Rob) and make this changes into a separate pull request.

(as a side note, we should also do something similar for the colors used in the browserAction popup).

Here is the diff of the changes I applied locally:

```diff
diff --git a/src/activitylog/activitylog.css b/src/activitylog/activitylog.css
index bfc5bf6..bfbf619 100644
--- a/src/activitylog/activitylog.css
+++ b/src/activitylog/activitylog.css
@@ -54,6 +54,7 @@ input[type='file'] {
   grid-column-gap: 10px;
   align-items: center;
   font-size: 14px;
+  border-bottom: 1px solid var(--table-splitter-color);
 }
 
 .options-dropdown {
@@ -152,8 +153,9 @@ input[type='file'] {
   height: 30px;
   align-items: center;
   padding: 0 15px;
-  background: #00c8d7;
-  color: #ffffff;
+  background: var(--table-header-background-color);
+  color: var(--text-color);
+  border-top: 1px solid var(--table-splitter-color);
 }
 
 .logs-counter {
diff --git a/src/activitylog/activitylog.html b/src/activitylog/activitylog.html
index 3e42c08..4463ed9 100644
--- a/src/activitylog/activitylog.html
+++ b/src/activitylog/activitylog.html
@@ -4,6 +4,7 @@
     <meta charset="UTF-8" />
     <meta name="viewport" content="width=device-width, initial-scale=1.0" />
     <title>Extension Activity Monitor</title>
+    <link rel="stylesheet" href="/colors.css" />
     <link rel="stylesheet" href="./activitylog.css" />
     <script src="activitylog.js" type="module"></script>
     <script
diff --git a/src/colors.css b/src/colors.css
new file mode 100644
index 0000000..3e377fa
--- /dev/null
+++ b/src/colors.css
@@ -0,0 +1,85 @@
+:root {
+  /*
+   * Photon Colors CSS Variables v3.3.2
+   * - Colors are taken from https://github.com/FirefoxUX/photon-colors/blob/master/photon-colors.css
+   *   by the Firefox DevTools and available in mozilla-central as devtools/client/themes/variables.css.
+   * - Plus a few unofficial colors: a few intermediary values (e.g. Blue 45),
+   *   and lighter variants for the dark theme (e.g. Red 20, Red 40).
+   */
+  --magenta-50: #ff1ad9;
+  --magenta-65: #dd00a9;
+  --magenta-70: #b5007f;
+
+  --purple-50: #9400ff;
+  --purple-60: #8000d7;
+
+  --blue-30: #75baff;
+  --blue-40: #45a1ff;
+  --blue-50: #0a84ff;
+  --blue-50-a30: rgba(10, 132, 255, 0.3);
+  --blue-55: #0074e8;
+  --blue-60: #0060df;
+  --blue-70: #003eaa;
+  --blue-80: #002275;
+
+  --teal-60: #00c8d7;
+  --teal-70: #008ea4;
+
+  --green-50: #30e60b;
+  --green-60: #12bc00;
+  --green-70: #058b00;
+
+  --yellow-50: #ffe900;
+  --yellow-60: #d7b600;
+  --yellow-65: #be9b00;
+  --yellow-70: #a47f00;
+  --yellow-80: #715100;
+
+  --red-20: #ffb3d2;
+  --red-40: #ff3b6b;
+  --red-50: #ff0039;
+  --red-60: #d70022;
+  --red-70: #a4000f;
+
+  --grey-10: #f9f9fa;
+  --grey-10-a15: rgba(249, 249, 250, 0.15);
+  --grey-10-a20: rgba(249, 249, 250, 0.2);
+  --grey-10-a25: rgba(249, 249, 250, 0.25);
+  --grey-10-a30: rgba(249, 249, 250, 0.3);
+  --grey-20: #ededf0;
+  --grey-25: #e0e0e2;
+  --grey-30: #d7d7db;
+  --grey-35: #c3c3c6;
+  --grey-40: #b1b1b3;
+  --grey-43: #a4a4a4;
+  --grey-45: #939395;
+  --grey-50: #737373;
+  --grey-55: #5c5c5f;
+  --grey-60: #4a4a4f;
+  --grey-70: #38383d;
+  --grey-80: #2a2a2e;
+  --grey-85: #1b1b1d;
+  --grey-90: #0c0c0d;
+  --grey-90-a05: rgba(12, 12, 13, 0.05);
+  --grey-90-a10: rgba(12, 12, 13, 0.1);
+  --grey-90-a15: rgba(12, 12, 13, 0.15);
+  --grey-90-a20: rgba(12, 12, 13, 0.2);
+  --grey-90-a30: rgba(12, 12, 13, 0.3);
+}
+
+/* Components theme CSS variables */
+:root {
+  --text-color: var(--grey-70);
+
+  --table-header-background-color: var(--grey-10);
+  --table-splitter-color: var(--grey-25);
+  --table-odd-rows: var(--grey-90-a05);
+
+  --selection-background: var(--blue-55);
+  --selection-color: white;
+
+  --checkbox-color: var(--blue-40);
+  --checkbox-focused-color: var(--blue-55);
+}
+
+
diff --git a/src/lib/web-component/filter-option/filter-option.css b/src/lib/web-component/filter-option/filter-option.css
index ba2b49f..1bb1209 100644
--- a/src/lib/web-component/filter-option/filter-option.css
+++ b/src/lib/web-component/filter-option/filter-option.css
@@ -37,12 +37,12 @@
 }
 
 .filter-checkbox input:checked ~ .select-indicator {
-  background: #00c8d7;
+  background: var(--checkbox-color)
 }
 
 .filter-checkbox:hover input:not([disabled]):checked ~ .select-indicator,
 .filter-checkbox input:checked:focus ~ .select-indicator {
-  background: #008ea4;
+  background: var(--checkbox-focused-color)
 }
 
 .select-indicator:after {
diff --git a/src/lib/web-component/log-view/log-view.css b/src/lib/web-component/log-view/log-view.css
index 74791ed..a6ee517 100644
--- a/src/lib/web-component/log-view/log-view.css
+++ b/src/lib/web-component/log-view/log-view.css
@@ -53,8 +53,8 @@ table {
 table thead {
   position: sticky;
   top: 0;
-  background: #00c8d7;
-  color: #ffffff;
+  background: var(--table-header-background-color);
+  color: var(--text-color);
 }
 
 table thead th {
@@ -62,6 +62,7 @@ table thead th {
   font-size: 16px;
   position: relative;
   text-align: left;
+  border: 1px solid var(--table-splitter-color);
 }
 
 table tbody {
@@ -71,7 +72,7 @@ table tbody {
 table thead th:not(:first-child)::after {
   content: '';
   display: block;
-  border: 1px solid #d7d7db;
+  border-right: 1px solid var(--table-splitter-color);
   position: absolute;
   top: 0;
   bottom: 0;
@@ -79,7 +80,7 @@ table thead th:not(:first-child)::after {
 }
 
 table tbody tr:nth-of-type(odd) {
-  background: #ededf0;
+  background: var(--table-odd-rows);
 }
 
 table tbody tr {
@@ -92,15 +93,16 @@ table td {
 }
 
 table td:not(:first-child) {
-  border-left: 1px solid #d7d7db;
+  border-left: 1px solid var(--table-splitter-color);
 }
 
 table td:not(:last-child) {
-  border-right: 1px solid #d7d7db;
+  border-right: 1px solid var(--table-splitter-color);
 }
 
 .row-highlight {
-  background-color: #96eaf0 !important;
+  background-color: var(--selection-background) !important;
+  color: var(--selection-color);
 }
 
 .table-empty-label {
```

src/lib/ext-activitylog.js
atiqueahmedziad: I have removed the method and I have added/removed CSS class to deal with showing/hiding dropdown now .

src/lib/ext-activitylog.js
atiqueahmedziad: Thanks for mentioning this. I have made it a singleton design. Let me know if I did it correctly.

src/lib/web-component/filter-option/filter-option-element.js
atiqueahmedziad: I used the `:empty` selector and pseudo-elements to fix this.

src/lib/web-component/filter-option/filter-option.css
atiqueahmedziad: Thanks a lot @rpl for explaining and for putting the diff. I have applied these styles.

src/lib/DropDownController.js
```diff
@@ -0,0 +1,54 @@
+class DropDownController {
+  constructor() {
```
Rob--W, line 2: Remove this whole constructor. It is not serving any purpose.

src/lib/DropDownController.js
```diff
@@ -0,0 +1,54 @@
...
+
+    DropDownController.instance = this;
+  }
+
+  toggleDropDown(showDropDown) {
+    if (this.elem?.toggleDropDown) {
+      this.elem.toggleDropDown(showDropDown);
+    } else {
+      const dropDown = this.elem.nextElementSibling;
+      showDropDown
```
Rob--W, line 15: Use classList.toggle - 
`dropDown.classList.toggle('expanded', showDropDown);`

src/lib/DropDownController.js
```diff
@@ -0,0 +1,54 @@
...
+    if (DropDownController.instance instanceof DropDownController) {
+      return DropDownController.instance;
+    }
+
+    DropDownController.instance = this;
+  }
+
+  toggleDropDown(showDropDown) {
+    if (this.elem?.toggleDropDown) {
+      this.elem.toggleDropDown(showDropDown);
```
Rob--W, line 12: Why are you still using `toggleDropDown`?

Correct me if I'm wrong, but it should be possible to achieve the desired behavior by passing the correct element to the `toggleDropDown` method. The rest of the styling can be taken care of by CSS.

src/lib/web-component/filter-option/filter-option-element.js
```diff
@@ -24,8 +26,10 @@ export class FilterOption extends HTMLElement {
     this.filterOptionTitle = filterContainer.querySelector('.title');
     this.filterOptionTitle.textContent = this.textContent;
 
-    this.checkboxList = filterContainer.querySelector('.checkbox-list');
-    this.checkboxList.classList.add(this.filterKey);
+    this.dropDownList = filterContainer.querySelector(
```
Rob--W, line 29: Why did you change the name? `checkboxList` is a better description of the functionality. The fact that it looks like a dropdown is not relevant.

src/lib/web-component/filter-timestamp/filter-timestamp-element.js
```diff
@@ -75,6 +84,12 @@ export class FilterTimestamp extends HTMLElement {
     }
   }
 
+  setTimestampStatus() {
+    if (this.timeStamp != null || Object.keys(this.timeStamp).length !== 0) {
+      this.classList.add('expanded');
```
Rob--W, line 89: This should go through the dropdowncontroller, shouldn't it?

When I set a time range, the dropdown appears.
When I open another dropdown or click anywhere else, the timestamp doesn't disappear.

src/lib/DropDownController.js
```diff
@@ -0,0 +1,35 @@
...
+  handleEvent(event) {
+    if (event.type === 'click' && event.target !== this.elem) {
+      this.hideDropDown();
+    }
+  }
+
+  triggerDropDown(elem) {
+    if (this.elem === elem) {
+      this.elem = null;
+      elem.classList.toggle('expanded', false);
```
Rob--W, line 11: Replace with `this.hideDropDown();`. The listener needs to be removed.

And perhaps the method should be renamed to `toggleDropDown` since you're toggling?

Rob--W (changes requested):

src/lib/web-component/filter-timestamp/filter-timestamp-element.js
atiqueahmedziad: Sorry, Yes, it is. I just saw it after pushing the last commit f275fef. Thanks for noticing.

src/lib/DropDownController.js
atiqueahmedziad: Yes, I have replaced this part with `this.hideDropDown()` method and renamed the `triggerDropDown` method to `toggleDropDown`

src/lib/web-component/filter-timestamp/filter-timestamp-element.js
```diff
@@ -75,6 +83,14 @@ export class FilterTimestamp extends HTMLElement {
     }
   }
 
+  setTimestampStatus() {
+    if (this.timeStamp != null || Object.keys(this.timeStamp).length !== 0) {
+      if (!this.classList.contains('expanded')) {
```
Rob--W, line 88: This "expanded" class name is an implementation detail of DropDownController. To avoid having to check the `expanded` class here, let `DropDownController` have three methods:

- showDropDown
- hideDropDown
- toggleDropDown.

... and call the right one from the right location.

Triggering a fake `click` event from here looks strange too. You should directly invoke the `dropDownController` instead.

src/lib/ext-activitylog.js
```diff
@@ -114,13 +115,20 @@ class Model {
...
+    this.clearLogBtn = document.querySelector('.clear-logs-btn');
+    this.optionsDropdown = document.querySelector('.options-dropdown');
+    this.saveLogBtn = document.querySelector('.save-log-btn');
     this.loadLogFile = document.querySelector('input[name="loadLogFile"]');
     this.notice = document.querySelector('.notice');
-    this.logHeading = document.querySelector('.log-heading');
+    this.filterIdTxt = document.querySelector('.filter-tabid');
+    this.pageType = document.querySelector('.page-type');
+    this.menuWrapper = document.querySelector('.menu-wrapper');
+    this.logCounter = document.querySelector('.logs-counter');
```
Rob--W, line 129: This member is not used anywhere else, either turn it into a variable, or remove it and use something like `.logs-counter > :last-child` (and `:first-child`) instead of the current two lines.

src/lib/ext-activitylog.js
```diff
@@ -226,8 +245,14 @@ class View {
     this.logView.clearTable();
   }
 
-  renderHeading({ tabId }) {
-    this.logHeading.textContent = `Activity Logs Filtered By Tab Id: ${tabId}`;
+  updatePageType(heading) {
```
Rob--W, line 248: This function is called only once, and its purpose in isolation is unclear. Let's remove the function and inline the logic at the caller.

src/lib/web-component/log-view/log-view-element.js
```diff
@@ -71,9 +73,6 @@ export class LogView extends HTMLElement {
       if (log.type === 'content_script') {
         const contentScriptTD = logTableRowInstance.querySelector('.name');
         contentScriptTD.textContent = log.name;
-        contentScriptTD.colSpan = 2;
```
Rob--W, line 74: Since we're no longer merging the columns, the whole logic here can be simplified by removing the `log.type` comparison. For example to

```javascript
      logTableRowInstance.querySelector('.api-name').textContent = log.name;
      logTableRowInstance.querySelector('.view-type').textContent = log.viewType || "";
```

Rob--W (commented):
I did some testing and did not find functional bugs; then I did one last pass at code review and found some points to address.

rpl:
@atiqueahmedziad the CI job is failing because of a failure in one of the unit tests: https://travis-ci.org/github/mozilla/extension-activity-monitor/builds/727874510#L259-L274

src/lib/web-component/filter-timestamp/filter-timestamp-element.js
atiqueahmedziad: Made these 3 methods in `DropDownController`. Used the `showDropDown` method in this case.

src/lib/ext-activitylog.js
atiqueahmedziad: turned it into local variable.

src/lib/ext-activitylog.js
atiqueahmedziad: Removed the function and put the logic at the caller.

src/lib/web-component/log-view/log-view-element.js
atiqueahmedziad: Thanks. I have updated this section.

tests/ext-activitylog.test.js
```diff
@@ -81,26 +81,16 @@ test('show/hide logs associated with extension id that is checked/unchecked from
...
-    '.checkbox-list'
-  );
-
   activityLog.handleNewLogs(logs);
 
   const tableBody = activityLog.view.logView.shadowRoot.querySelector('tbody');
   const getTableRows = () => tableBody.querySelectorAll('tr');
   let tableRows = getTableRows();
 
-  // Initially filter Checkbox List is hidden
```
Rob--W, line 94: Instead of just removing it, it would be good to actually verify that the dropdown logic works as expected.

atiqueahmedziad: added the test to verify the dropdown logic.

rpl:
lgtm too, we can defer more changes to separate PR.

As a side note, while trying this locally I noticed an issue when monitoring extensions that subscribes the tabs.onUpdated API event, I filed an issue with some more details and the steps to reproduce it as #57.

rpl (approved):

PR #54, opened 2020-09-15
feat: logs counter added

### LINKED ISSUE #32: Log counter should be displayed

atiqueahmedziad:
Blocks #52 
Fixes #32

### DIFF
--- a/src/lib/web-component/log-view/log-view-element.js
+++ b/src/lib/web-component/log-view/log-view-element.js
@@ -3,6 +3,8 @@ import { dateTimeFormat } from '../../formatters.js';
 export class LogView extends HTMLElement {
   constructor() {
     super();
+
+    this.totalLogs = 0;
     this.isFilterMatched = () => true;
 
     const shadow = this.attachShadow({ mode: 'open' });
@@ -27,10 +29,19 @@ export class LogView extends HTMLElement {
 
   setLogFilter(filterFunc) {
     this.isFilterMatched = filterFunc;
+    let filteredLogs = 0;
 
     for (const row of this.tableBody.rows) {
       row.hidden = !this.isFilterMatched(row._log);
+
+      if (!row.hidden) {
+        filteredLogs++;
+      }
     }
+
+    this.dispatchEvent(
+      new CustomEvent('logcounter', { detail: { filteredLogs } })
+    );
   }
 
   addNewRows(logs) {
@@ -64,6 +75,11 @@ export class LogView extends HTMLElement {
       rowsFragment.appendChild(logTableRowInstance);
     }
     this.tableBody.appendChild(rowsFragment);
+
+    this.totalLogs += logs.length;
+    this.dispatchEvent(
+      new CustomEvent('logcounter', { detail: { totalLogs: this.totalLogs } })
+    );
   }
 
   handleEvent(event) {
--- a/src/lib/web-component/log-view/log-view-element.js
+++ b/src/lib/web-component/log-view/log-view-element.js
@@ -29,18 +29,26 @@ export class LogView extends HTMLElement {
 
   setLogFilter(filterFunc) {
     this.isFilterMatched = filterFunc;
-    let filteredLogs = 0;
 
     for (const row of this.tableBody.rows) {
       row.hidden = !this.isFilterMatched(row._log);
+    }
 
+    this.updateFilteredRowsCount();
+  }
+
+  updateFilteredRowsCount() {
+    let visibleRows = 0;
+    for (const row of this.tableBody.rows) {
       if (!row.hidden) {
-        filteredLogs++;
+        visibleRows++;
       }
     }
 
     this.dispatchEvent(
-      new CustomEvent('logcounter', { detail: { filteredLogs } })
+      new CustomEvent('logcountchange', {
+        detail: { filteredLogs: visibleRows },
+      })
     );
   }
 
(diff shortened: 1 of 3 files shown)


src/lib/web-component/log-view/log-view-element.js
```diff
@@ -27,10 +29,19 @@ export class LogView extends HTMLElement {
...
     for (const row of this.tableBody.rows) {
       row.hidden = !this.isFilterMatched(row._log);
+
+      if (!row.hidden) {
+        filteredLogs++;
+      }
     }
+
+    this.dispatchEvent(
+      new CustomEvent('logcounter', { detail: { filteredLogs } })
```
Rob--W, line 43: `logcountchange` would be a better event name than "logcounter".

src/lib/web-component/log-view/log-view-element.js
```diff
@@ -64,6 +75,11 @@ export class LogView extends HTMLElement {
       rowsFragment.appendChild(logTableRowInstance);
     }
     this.tableBody.appendChild(rowsFragment);
+
+    this.totalLogs += logs.length;
+    this.dispatchEvent(
+      new CustomEvent('logcounter', { detail: { totalLogs: this.totalLogs } })
```
Rob--W, line 81: This logic is incorrect, you also need to emit the number of filtered logs, as that may change by the outcome of `logTableRowInstance.hidden = !this.isFilterMatched(log);` above.

Rob--W (changes requested):

src/lib/web-component/log-view/log-view-element.js
atiqueahmedziad: aren't we showing the total number of logs that are collected in the background here?

Rob--W: We are showing the total number of logs, whether it's from the background or a file doesn't matter.

src/lib/web-component/log-view/log-view-element.js
atiqueahmedziad: yes, it is. Updated to the event name to `logcountchange`. :-)

src/lib/web-component/log-view/log-view-element.js
atiqueahmedziad: I made another method named `updateFilteredRowsCount` which gets called when filters are applied OR when new logs are rendered.

src/lib/web-component/log-view/log-view-element.js
```diff
@@ -64,6 +83,15 @@ export class LogView extends HTMLElement {
...
     }
     this.tableBody.appendChild(rowsFragment);
+
+    this.updateFilteredRowsCount();
+
+    this.totalLogs += logs.length;
+    this.dispatchEvent(
+      new CustomEvent('logcountchange', {
+        detail: { totalLogs: this.totalLogs },
+      })
```
Rob--W, line 93: Merge this logic with `updateFilteredRowsCount`.

Rob--W (changes requested):

src/lib/web-component/log-view/log-view-element.js
atiqueahmedziad: I renamed the method to `updateRowCountAndStyle`. Any suggestion?

src/lib/web-component/log-view/log-view-element.js
```diff
@@ -3,6 +3,8 @@ import { dateTimeFormat } from '../../formatters.js';
 export class LogView extends HTMLElement {
   constructor() {
     super();
+
+    this.totalLogs = 0;
```
Rob--W, line 7: This member isn't needed, it can be a local variable.

src/lib/web-component/log-view/log-view-element.js
```diff
@@ -31,6 +33,23 @@ export class LogView extends HTMLElement {
     for (const row of this.tableBody.rows) {
       row.hidden = !this.isFilterMatched(row._log);
     }
+
+    this.updateRowCountAndStyle();
+  }
+
+  updateRowCountAndStyle() {
```
Rob--W, line 40: Let's change it to `triggerLogCountChange()`.

I guess that the `AndStyle` suffix was meant to account for future changes to the even-odd appearance.

Given the potential performance impact of the current implementation (see issue 55), I think that it may make sense to remove filtered rows from the document instead.

src/lib/web-component/log-view/log-view-element.js
atiqueahmedziad: Made it local variable.

src/lib/web-component/log-view/log-view-element.js
atiqueahmedziad: Renamed the method to `triggerLogCountChange`

src/lib/web-component/log-view/log-view-element.js
```diff
@@ -1,5 +1,7 @@
 import { dateTimeFormat } from '../../formatters.js';
 
+let totalLogs = 0;
```
Rob--W, line 3: This is not a local variable but a global variable. The value is only used inside `triggerLogCountChange`, so you may as well compute the value from there (`this.tableBody.rows.length` at the moment).

Rob--W (approved):
This patch in isolation doesn't do anything. It would have been nice to have added unit tests to verify that the logic is correct.

I'm not going to block on that though, since you've mention that you're going to use this in #52.

PR #36, opened 2020-08-07
Curves at edges

himanshujaidka:
Screenshot (31)
A curvy button from edges looks like more interesting

PR #61, opened 2020-09-29
I have made a little changes to the button of popup

himanshujaidka:
Screenshot (85)


wagnerand:
I am not a designer but I don't think we should add a border radius. As you can see from your screenshot, no other elements has one, even the first button in the popup wouldn't have it with this patch. Also, nowhere in Firefox is that border radius used, so it looks a bit out of place.

himanshujaidka:
okay

cadeyrn:
by the way, it's exactly the same PR as already in #35 and in #36…

PR #76, opened 2020-10-01
fix: Re-styling browser action popup using browser_style and photon colors

Parimalisgreat:
Fix #76 
improved popup.css browserAction popup


src/popup/popup.css
```diff
@@ -20,15 +20,17 @@ button {
   font-size: 14px;
   color: #000000;
   background-color: #ffffff;
-  border: 1px solid #000000;
+  /* border: 1px solid #000000; */
```
rpl, line 23: Nit, in general it is better to do not comment out the parts that you want to replace (1), just remove them, the diff will already tell us what was the patch replaced.

----

(1): in this case it is easy to spot them, but in a bigger diff some of the commented parts may go unnoticed and may land on master by mistake

src/popup/popup.css
```diff
@@ -60,12 +67,13 @@ button#stopMonitorBtn:hover {
 }
 
 .success {
-  background-color: #12bc00;
+  background-color: rgb(9, 92, 9);
```
rpl, line 70: There is colors.css file with CSS variables for the colors from the photon style guide, we should include that stylesheet file and replace the colors used in the popup css with CSS variables (not just the colors being changed in this PR, also the other already being used).

The colors.css file is currently in the `src/activitylog/` directory because at the moment it is only used in the activitylog.html, we may move it in a shared stylesheet directory now that the popup should also be using it.

src/popup/popup.css
```diff
@@ -20,15 +20,17 @@ button {
   font-size: 14px;
```
rpl, line 20: The photon style guide also provide some guidance related to the size, spacing and font size of the common UI components, like buttons. Let's double-check those and try to match the suggestions, here is the page that describes the photon buttons styles: https://design.firefox.com/photon/components/buttons.html

src/popup/popup.css
```diff
@@ -60,12 +67,13 @@ button#stopMonitorBtn:hover {
...
 .success {
-  background-color: #12bc00;
+  background-color: rgb(9, 92, 9);
   color: #ffffff;
 }
 
 .failure,
 #errorText {
   background-color: #d70022;
+
```
rpl, line 77: Nit, remove changes unrelated to the issue being fixed, e.g. this new empty line here and in the popup.html (especially if there isn't any particular reason to apply that change). In general the patch diff should be as focused on the issue being fixed (and clean) as possible.

rpl (changes requested):
@Parimalisgreat thanks for looking into fixing this issue, follows an initial round of review comments.

Besides that, given the change applied are related to the popup UI styles, it would be helpful to include in the issue description a small screenshot of the popup before and after the changes applied in the PR, so that we can get an idea (even before we may be pulling the PR locally and giving it a try).

wagnerand:
@Parimalisgreat Hey, can you please reference the issue this PR is supposed to fix in your first comment like so: "Fix #<issuenumber>"? Thanks.

Parimalisgreat:
@wagnerand @rpl thanks for looking into my PR i had made all the suggested changes. Please inform me if any further changes are required . I am soon going to make a PR . Below is the screenshot of the UI .
Capture

src/popup/popup.css
```diff
@@ -18,40 +18,44 @@ button {
   height: 35px;
   width: 100%;
   font-size: 14px;
-  color: #000000;
-  background-color: #ffffff;
-  border: 1px solid #000000;
+  color: var(--grey-90-a30);
+  background-color: var(--selection-color);
```
rpl, line 22: Follows some additional review comments related to the usage of the CSS variables:

- instead of using directly a CSS variable related to a particular color (like `var(--grey-90-a30)`), we should define variables with names associated to what they are used for, as an example:
  - in the photon style guide section that I linked in my previous comment (https://design.firefox.com/photon/components/buttons.html#styles) there are two possible button styles the primary one (which is blue) and the default one (which is grey) and so we may define two variables named `var(--button-default-background-color)` and `var(--button-primary-background-color)` and set them to their default photon color (not arbitrary ones, the ones that are explicitly mentioned in the guide)

- `--selection-color` is currently being used for the styling related to a selected table row, it doesn't really match what it is being used for here, where it is set as the background color of a button, we shouldn't use CSS variables that have a different meaning because it would be confusing, as also described in the previous point we should define [new CSS variables in the section here](https://github.com/mozilla/extension-activity-monitor/blob/5e609b6c2deb2acc7e41d521e8f469bc379164dd/src/activitylog/colors.css#L71-L84) when there isn't already one that match what we want to use the color for.

Besides that, based on the role of the buttons in the popup:

- the "Start monitoring ..." or "Stop monitoring ..." ([1]) button that is currently active should probably be styled as a primary button based on the photon style guide, the inactive one should be styled as an inactive button

- the "View activity log page" button does not seem to be a primary button and so it should be styled with the default button style (grey), purple seems an arbitrary color for it

The buttons also do not yet look as Photon styled buttons, if we want to follow the photon style guide a bit more we should also follow the styling rules described here: https://design.firefox.com/photon/components/buttons.html#sizes 

Same comments are valid for the other colors associated to other components in the popup, e.g. the element with the id `monitorStatus` looks a lot like a message bar (https://design.firefox.com/photon/components/message-bars.html), I think that being red when the monitor is not active is a bit extreme because it is not an error state, it could instead be styled as a `"notice" message bar` or maybe a `"warning" message bar`, because we just want to make the user aware that the monitor is not active yet (on the contrary when the monitor becomes active it seems reasonable that it would look like a `"success" message bar`, which is green in the photon style guide). Also in this case we should create CSS variables associated to the the role instead of the current color (like `--messagebar-notice-background-color`, `--messagebar-warning-background-color` etc.) 

----

(1): it may be actually make sense to merge the two Start and Stop monitor buttons into a single button that change state (instead of two separate buttons), but we can defer that to a separate follow up

rpl (changes requested):

wagnerand:
@Parimalisgreat We haven't seen much activity in a long time, so we are closing this PR. Feel free to reopen it if you resume work on it.

PR #88, opened 2020-10-07
fix: close log details sidebar along with clearing all logs in the log-view table.

### LINKED ISSUE #56: Log details still showing details after clearing the logs
The log details view is still showing a log even after the clear button was clicked to clear all logs.

I expected the details view to have been cleared together with the log view table.

himanshujaidka:
Fixes #56

### DIFF
--- a/src/lib/ext-activitylog.js
+++ b/src/lib/ext-activitylog.js
@@ -244,7 +244,6 @@ class View {
 
   clearTable() {
     this.logView.clearTable();
-    this.logView.closeDetailSidebar();
   }
 }
 
--- a/src/lib/ext-activitylog.js
+++ b/src/lib/ext-activitylog.js
@@ -244,6 +244,7 @@ class View {
 
   clearTable() {
     this.logView.clearTable();
+    this.logView.closeDetailSidebar();
   }
 }
 
--- a/src/lib/web-component/log-view/log-view-element.js
+++ b/src/lib/web-component/log-view/log-view-element.js
@@ -136,6 +136,7 @@ export class LogView extends HTMLElement {
   clearTable() {
     this.tableBody.textContent = '';
     this.emptyTableLabel.hidden = false;
+    this.closeDetailSidebar();
   }
 
   connectedCallback() {
--- a/tests/ext-activitylog.test.js
+++ b/tests/ext-activitylog.test.js
@@ -282,7 +282,14 @@ test('clearing logs from activitylog page', async () => {
   expect(activityLog.model.logs).toMatchObject(logs);
   expect(getTableRows().length).toBe(2);
 
+  const [row] = getTableRows();
+  const { logDetailWrapper } = activityLog.view.logView;
+  expect(logDetailWrapper.hidden).toBe(true);
+  row.click();
+  expect(logDetailWrapper.hidden).toBe(false);
+
   clearLogBtn.click();
+  expect(logDetailWrapper.hidden).toBe(true);
 
   expect(clearBackgroundLogsFn).toHaveBeenCalled();
   expect(activityLog.model.logs).toMatchObject([]);


src/lib/ext-activitylog.js
```diff
@@ -244,6 +244,7 @@ class View {
 
   clearTable() {
     this.logView.clearTable();
+    this.logView.closeDetailSidebar();
```
rpl, line 247: This is still the same change that was in #86 (just the unrelated empty line change has been removed from what was in #86), please reread the previous feedback comments and update this PR accordingly.

rpl (changes requested):
I did already explicitly mention in my previous review comments here https://github.com/mozilla/extension-activity-monitor/pull/86#issuecomment-704884177 and in the issue here https://github.com/mozilla/extension-activity-monitor/issues/56#issuecomment-704432760 what would be my preferred place and approach to fix the issue, please take the time to read more carefully those two comments and then update this PR accordingly. 

As an additional note:

do you think that the pull request description is now in sync with what the PR is doing?
is any part of it that is actually misleading and describing something that this PR isn't actually doing?

I'm not going to comment yet about the testing part (because it would be good to be sure that the change is actually applied as suggested in the previous feedback and it currently isn't) but, once that single line change is finally in the right place, additional test coverage will be needed to complete this PR.

src/lib/ext-activitylog.js
himanshujaidka: Okay sir, I will make changes in log-view-element.js according to you and thanks

himanshujaidka: sir, this pr is not eligible?

rpl: yes, I described the reason why we should be fix the issue in another source file (in `log-view-element.js` instead of `ext-activitylog.js`) in my previous feedback comments, eg. as I did mention in the issue here https://github.com/mozilla/extension-activity-monitor/issues/56#issuecomment-704432760:
> The log details sidebar is part (and managed by) the webcomponent defined in log-view-element.js.

In other words: given that there shouldn't be any real scenario that would require to clear the table but keep the sidebar open, we can assign the responsibility to close the sidebar automatically to the log-view webcomponent itself and make it something that happens internally in the `log-view` webcomponent when `logView.clearTable()` has been called.

have you tried to think / look into how the same change you applied here could be moved in `log-view-element.js`?

As you can see from this diff, both `clearTable` and `closeDetailSidebar` are methods of the `log-view` webcomponent defined in `log-view-element.js`.

From inside the [`clearTable` method defined in `log-view-lement.js`](https://github.com/mozilla/extension-activity-monitor/blob/a981ef40ccf2fa41fdc68fcddd0af771a1b43297/src/lib/web-component/log-view/log-view-element.js#L136-L139) we can call `this.closeDetailsSidebar()` and avoid any change to ext-activitylog.js.

(as a side effect of this choice, this behavior would be provided by the log-view component itself and so it would be supported even if we do reuse the log-view webcomponent into some other extension page that it isn't managed by the MVC implemented in ext-activitylog.js).

rpl (changes requested):

himanshujaidka:
@rpl sir, yes I understood your previous comment that `closeDetailSidebar` and `clearTable` both are defined inside the `log-view` component, in simple words, to run this extension smoothly we don't have to take the changes into `ex-activity`, so all the changes regarding sidebar should be there in `log-view` component and I have read all the codebase and understood it

rpl (changes requested):
> @rpl sir, yes I understood your previous comment that `closeDetailSidebar` and `clearTable` both are defined inside the `log-view` component, in simple words, to run this extension smoothly we don't have to take the changes into `ex-activity`, so all the changes regarding sidebar should be there in `log-view` component and I have read all the codebase and understood it

ok, we finally got the right fix in place, that's good but as I anticipated this pull request isn't really completed until we make sure that the tests do cover the fix.

For "the tests do cover the fix" I don't mean "the tests are all green", what it really means is "the tests do fail without the fix and pass with the fix in place", in other words the main point of writing test is that they fail if something breaks (if they don't... well they are not really useful)

At the moment the tests are all passing with and without the fix and so it is clear that the fix isn't covered.

To cover it ~10 new lines in an existing test case should be enough (:warning: **you do not need to move any other code around or update npm packages** :warning:), in particular in the existing 'clearing logs from activitylog page' test case ([defined in ext-activitylog.test.js here](https://github.com/mozilla/extension-activity-monitor/blob/a981ef40ccf2fa41fdc68fcddd0af771a1b43297/tests/ext-activitylog.test.js#L227)) we are already testing the behavior on clicking the "clear logs" button:

https://github.com/mozilla/extension-activity-monitor/blob/a981ef40ccf2fa41fdc68fcddd0af771a1b43297/tests/ext-activitylog.test.js#L283-L285

And so what we would like to assert around that is something like:

```js
expect(getTableRows().length).toBe(2);

// TODO 1: get a reference to one of the row elements
// TODO 2: assert that the sidebar is hidden
// TODO 3: click the row element
// TODO 4: assert that the sidebar is not hidden

 clearLogBtn.click();

// TODO 5: assert that the sidebar is hidden
```
Here are some pointers to achieve the todos (1) and (2):

1. to get a reference to one of the row element we can reuse the `getTableRows()` helpers (e.g. `const [row] = getTableRows();`)
2. to assert that the sidebar is hidden we have to retrieve a reference to it from `logView` (e.g. `const { logDetailWrapper } = activityLog.view.logView;`) and then assert that its `hidden` property is `true` (e.g. `expect(logDetailWrapper.hidden).toBe(true);`) 

And the other todo items should be easy enough to guess based on this pointers.

Try to run the tests as it is before apply your changes, to be sure that you can run it successfully before changing it, then add the additional tweaks as suggested above and try to rerun again.

Once you got the test to pass as expected, try to temporarily comment out the fix you applied in log-view-element.js and run the test again to be sure that it does fail as expected.

rpl:
It looks that you missed to notice (and answer) to this question from my previous review comments (https://github.com/mozilla/extension-activity-monitor/pull/88#pullrequestreview-504113778)

> do you think that the pull request description is now in sync with what the PR is doing?
> is any part of it that is actually misleading and describing something that this PR isn't actually doing?

please re-read the pull request title carefully, is there any part of `"fix: close log details sidebar along with clearing all logs in the log-view table and update dependencies"` that we can remove because it isn't really part of this pull request?

himanshujaidka:
> It looks that you missed to notice (and answer) to this question from my previous review comments ([#88 (review)](https://github.com/mozilla/extension-activity-monitor/pull/88#pullrequestreview-504113778))
> 
> > do you think that the pull request description is now in sync with what the PR is doing?
> > is any part of it that is actually misleading and describing something that this PR isn't actually doing?
> 
> please re-read the pull request title carefully, is there any part of `"fix: close log details sidebar along with clearing all logs in the log-view table and update dependencies"` that we can remove because it isn't really part of this pull request?
ok, sir I will look into it

himanshujaidka:
ok sir I will look into it

himanshujaidka:
sir, bi-mistake I have added the wrong title to this, sorry for that

himanshujaidka:
@rpl sir, pls have a look into it, I think I have made it now

rpl (approved):
lgtm, thanks

himanshujaidka:
Thanks, everyone, it was a great experience, I have learned new things in this pull request, again thank you to everyone, and looking forward to work under your guidance :)

himanshujaidka:
@rpl sir, would you mind if I am requesting you to add `hacktoberfest-accepted` label to this pr 😅?

caitmuenster:
Hi @himanshujaidka, we've actually decided to not participate in Hacktoberfest this year. Please find another project with that topic and contribute there.

himanshujaidka:
@cadecairos-mozilla-owner, ok mam no worries, thanks for the reply 😇

PR #152, opened 2021-01-06
chore(ci): Migrate CI jobs to circleci

rpl:
- migrate CI jobs from travis to circleci
- migrate code coverage reporting from coveralls to codecov


Rob--W (approved):
This looks so much more verbose than before. Anyway, rubberstamp lgtm.

rpl:
> This looks so much more verbose than before. Anyway, rubberstamp lgtm.

yeah, that's a bit unfortunate, I initially derived it from some of the other circleci configs, then I simplified it a bit given that this project doesn't need most of what we needed in some of the other repos (e.g. this repo doesn't need to release a package on npm, or run the tests with multiple nodejs versions).

I could tweak it further and make it even a bit simpler (e.g. by removing the node_modules caching, remove the references and parameters may also reasonable with what the workflow currently does), but I wasn't sure it is really worth it. We don't change the ci jobs that often and so it may not be such a big deal.

PR #332, opened 2022-09-15
refactor: browserAction popup UI tests improved

atiqueahmedziad:
Part of fix for #50 
### Fixes
1. Refactored the browserAction popup test (also did some refactoring in `ext-popup.js`) with meaningful tests.

### DIFF
--- a/src/lib/constant.js
+++ b/src/lib/constant.js
@@ -1,0 +1,2 @@
+export const MONITOR_ENABLED_MSG = 'Extensions are being monitored';
+export const MONITOR_DISABLED_MSG = 'No extensions are being monitored';
--- a/src/lib/ext-popup.js
+++ b/src/lib/ext-popup.js
@@ -1,3 +1,4 @@
+import { MONITOR_DISABLED_MSG, MONITOR_ENABLED_MSG } from './constant.js';
 import {
   getMonitorStatus,
   startMonitor,
@@ -23,7 +24,7 @@ export default class Popup {
     this.startMonitorAllBtn.setAttribute('disabled', true);
     this.stopMonitorAllBtn.removeAttribute('disabled');
 
-    this.monitorStatusText.textContent = 'Extensions are being monitored';
+    this.monitorStatusText.textContent = MONITOR_ENABLED_MSG;
     this.monitorStatusText.classList.add('success');
     this.monitorStatusText.classList.remove('failure');
   }
@@ -32,7 +33,7 @@ export default class Popup {
     this.startMonitorAllBtn.removeAttribute('disabled');
     this.stopMonitorAllBtn.setAttribute('disabled', true);
 
-    this.monitorStatusText.textContent = 'No extensions are being monitored';
+    this.monitorStatusText.textContent = MONITOR_DISABLED_MSG;
     this.monitorStatusText.classList.add('failure');
     this.monitorStatusText.classList.remove('success');
   }
@@ -44,33 +45,25 @@ export default class Popup {
       : this.renderMonitorStoppedUI();
   }
 
-  handleViewActivityLog() {
-    openActivityLogPage();
+  async handleViewActivityLog() {
+    await openActivityLogPage();
     window.close();
   }
 
   async handleEvent(event) {
     try {
-      if (event.type === 'click') {
-        switch (event.target) {
-          case this.startMonitorAllBtn:
-            await this.handleMonitor(startMonitor);
-            break;
-          case this.stopMonitorAllBtn:
-            await this.handleMonitor(stopMonitor);
-            break;
-          case this.openActivityLogBtn:
-            this.handleViewActivityLog();
-            break;
-          default:
-            throw new Error(
-              'wrong event target id found: ' + JSON.stringify(event.target.id)
-            );
-        }
-      } else {
-        throw new Error(
-          'wrong event type found: ' + JSON.stringify(event.type)
-        );
+      switch (event.target) {
+        case this.startMonitorAllBtn:
+          await this.handleMonitor(startMonitor);
+          break;
+        case this.stopMonitorAllBtn:
+          await this.handleMonitor(stopMonitor);
+          break;
+        case this.openActivityLogBtn:
+          await this.handleViewActivityLog();
+          break;
+        default:
+          break;
       }
     } catch (error) {
       this.renderErrorMsg(error.message);
@@ -78,12 +71,8 @@ export default class Popup {
   }
 
   async handleMonitor(monitorFunc) {
-    try {
-      await monitorFunc();
-      this.render();
-    } catch (error) {
-      this.renderErrorMsg(error.message);
-    }
+    await monitorFunc();
+    this.render();
   }
 
   async init() {
--- a/src/lib/ext-popup.js
+++ b/src/lib/ext-popup.js
@@ -63,7 +63,9 @@ export default class Popup {
           await this.handleViewActivityLog();
           break;
         default:
-          break;
+          throw new Error(
+            'wrong event target id found: ' + JSON.stringify(event.target.id)
+          );
       }
     } catch (error) {
       this.renderErrorMsg(error.message);
--- a/tests/ext-listen.test.js
+++ b/tests/ext-listen.test.js
@@ -47,17 +47,3 @@ test('startMonitor and stopMonitor should have their respective requestType', as
     requestType: 'stopMonitor',
   });
 });
-
-test('openActivityLogPage should create tab with ActivityLogPageURL', async () => {
-  const getURL = jest.fn();
-  const create = jest.fn();
-
-  window.browser = {
-    runtime: { getURL },
-    tabs: { create },
-  };
-
-  ExtListen.openActivityLogPage();
-  expect(create).toHaveBeenCalled();
-  expect(getURL).toHaveBeenCalled();
-});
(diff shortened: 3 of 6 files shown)


tests/ext-listen.test.js
```diff
@@ -47,17 +47,3 @@ test('startMonitor and stopMonitor should have their respective requestType', as
     requestType: 'stopMonitor',
   });
 });
-
-test('openActivityLogPage should create tab with ActivityLogPageURL', async () => {
```
atiqueahmedziad, line 51: This has got coverage while testing the opening of Activity Logs page from browerAction popup (in `ext-popup.test.js`). Hence removing this.

src/lib/ext-popup.js
```diff
@@ -44,46 +45,34 @@ export default class Popup {
...
-  handleViewActivityLog() {
-    openActivityLogPage();
+  async handleViewActivityLog() {
+    await openActivityLogPage();
     window.close();
   }
 
   async handleEvent(event) {
     try {
-      if (event.type === 'click') {
```
atiqueahmedziad, line 54: Removed this condition since we are have only attached [click event on the buttons](https://github.com/mozilla/extension-activity-monitor/blob/master/src/lib/ext-popup.js#L91-L93). So, the else call is unreachable in real scenario.

src/lib/ext-popup.js
```diff
@@ -44,46 +45,36 @@ export default class Popup {
...
   }
 
   async handleMonitor(monitorFunc) {
-    try {
-      await monitorFunc();
-      this.render();
-    } catch (error) {
-      this.renderErrorMsg(error.message);
-    }
+    await monitorFunc();
```
Rob--W, line 76: Why did we have try-catch before, and how do we know that removing try-catch is fine?

src/lib/ext-popup.js
Rob--W: To ensure that this continues the case, consider adding an assertion check with
https://developer.mozilla.org/en-US/docs/Web/API/Console/assert

tests/ext-listen.test.js
Rob--W, line 51: Why was this test removed?

src/lib/constant.js
```diff
@@ -0,0 +1,2 @@
+export const MONITOR_ENABLED_MSG = 'Extensions are being monitored';
```
Rob--W, line 1: Why have these strings become constants? This introduces some inconsistency: lots of UI strings are embedded in the code itself, except for some strings.

It looks like this was used in tests; in that case you could keep the original implementation as is, and declare constants in the test with a comment before pointing out that this is a string in ext-popup.js

Rob--W (commented):
I took a very quick look. I didn't look at ext-popup.test.js because it wasn't clear what you've been trying to do there. This PR says that the "popup UI tests improved", but I other than it being a refactor it's not clear in what way the code has been improved. Could you elaborate?

tests/ext-popup.test.js
```diff
@@ -23,187 +26,205 @@ function observeChange(element) {
...
+const startMonitorSpyFn = jest.spyOn(ExtListen, 'startMonitor');
+const stopMonitorSpyFn = jest.spyOn(ExtListen, 'stopMonitor');
+
+afterEach(() => {
+  getMonitorStatusSpyFn.mockReset();
+  startMonitorSpyFn.mockReset();
+  stopMonitorSpyFn.mockReset();
+});
+
+test('clicking on start monitoring button should display that the extensions are being monitored', async () => {
```
atiqueahmedziad, line 39: This description - `clicking on start monitoring button should display that the extensions are being monitored` is more aligned with what a user should expect after clicking the start monitoring button rather than a description like - `start monitoring all extensions from the popup`. So, I believe a new developer will find the new test name more easily understandable.

tests/ext-popup.test.js
```diff
@@ -23,187 +26,205 @@ function observeChange(element) {
...
   document.body.innerHTML = popupBody;
 
-  const getMonitorStatusFn = jest.spyOn(ExtListen, 'getMonitorStatus');
-  const startMonitorFn = jest.spyOn(ExtListen, 'startMonitor');
+  // no extensions are being monitored initially
```
atiqueahmedziad, line 42: Added comment here as the line - 
```js
getMonitorStatusSpyFn.mockResolvedValueOnce(false);
```
doesn't explains what's happening if it's being resolved with boolean value. 
This improves the readability of this part.

tests/ext-popup.test.js
```diff
@@ -23,187 +26,205 @@ function observeChange(element) {
...
-  await popup.init();
-
-  // popup at initial - no extensions being watched
-  expect(startBtn.hasAttribute('disabled')).toBeFalsy();
-  expect(stopBtn.hasAttribute('disabled')).toBeTruthy();
-
-  expect(monitorStatus.textContent).toBe('No extensions are being monitored');
-  expect(monitorStatus.classList.contains('failure')).toBeTruthy();
-  expect(monitorStatus.classList.contains('success')).toBeFalsy();
-  expect(errorText.textContent).toBe('');
```
atiqueahmedziad, line 52: The initial UI behaviour of the browserAction popup is not necessary to be checked for the purpose of the test - `clicking on start monitoring button should display that the extensions are being monitored`.

Rather, I have added - 
```js
  // no extensions are being monitored initially
  getMonitorStatusSpyFn.mockResolvedValueOnce(false);
```
above which explains the initial state of the browserAction popup.
I think this improves the test's quality.

tests/ext-popup.test.js
```diff
@@ -23,187 +26,205 @@ function observeChange(element) {
...
-
-  expect(startBtn.hasAttribute('disabled')).toBeTruthy();
-  expect(stopBtn.hasAttribute('disabled')).toBeFalsy();
-
-  expect(monitorStatus.textContent).toBe('Extensions are being monitored');
-  expect(monitorStatus.classList.contains('failure')).toBeFalsy();
-  expect(monitorStatus.classList.contains('success')).toBeTruthy();
-
-  expect(errorText.textContent).toBe('');
-
```
atiqueahmedziad, line 100: The initial UI behaviour of the browserAction popup is not necessary to be checked for the purpose of the test - `clicking on stop monitoring button should display that the extensions are not being monitored`.

Rather, I have added - 
```js
  // assume, extensions are being monitored
  getMonitorStatusSpyFn.mockResolvedValueOnce(true);
```
above which explains the initial state of the browserAction popup.

tests/ext-popup.test.js
```diff
@@ -23,187 +26,205 @@ function observeChange(element) {
...
 
-  expect(monitorStatus.textContent).toBe('No extensions are being monitored');
+  expect(monitorStatus.textContent).toBe(MONITOR_DISABLED_MSG);
   expect(monitorStatus.classList.contains('failure')).toBeTruthy();
   expect(monitorStatus.classList.contains('success')).toBeFalsy();
   expect(errorText.textContent).toBe('');
+});
 
-  getMonitorStatusFn.mockClear();
+test('clicking on "view activity logs page" button should open a new tab with activity logs page', async () => {
```
atiqueahmedziad, line 101: This test name is improved from `clicking on "view activity logs page" button should open the activitylog page` to `clicking on "view activity logs page" button should open a new tab with activity logs page`, which seemed more concrete result that we expect.

tests/ext-popup.test.js
```diff
@@ -23,187 +26,205 @@ function observeChange(element) {
...
+  expect(getURL).toHaveBeenCalled();
+  expect(create).toHaveBeenCalledWith({ url: ACTIVITY_LOG_PAGE_URL });
+
+  setImmediate(() => {
+    // expect the browserAction popup to autoclose itself.
+    expect(window.close).toHaveBeenCalled();
+  });
 });
 
-test('errors on start/stopping monitoring should be shown in the popup UI', async () => {
```
atiqueahmedziad, line 116: this test has been split into two tests, one is  `error at start monitoring extensions should be displayed` and another one is `error at stop monitoring extensions should be displayed`. This made the test code more aligned for its single purpose.

tests/ext-popup.test.js
```diff
@@ -23,187 +26,205 @@ function observeChange(element) {
...
   stopBtn.click();
-  await promiseErrorChanged2;
+  await observeErrorTextChange;
 
-  expect(errorText.textContent).toBe('stop-monitor-error');
+  expect(errorText.textContent).toBe(STOP_MONITOR_ERROR);
 });
 
-test('Error is shown in popup UI while wrong parameter is passed to handleEvent', async () => {
+test('error at the opening of activity log page should be displayed', async () => {
```
atiqueahmedziad, line 180: This test was missing. I have added this test.

atiqueahmedziad (commented):
> I took a very quick look. I didn't look at ext-popup.test.js because it wasn't clear what you've been trying to do there. This PR says that the "popup UI tests improved", but I other than it being a refactor it's not clear in what way the code has been improved. Could you elaborate?

Sorry for not elaborating this in the description. The tests are improved by doing the follows -
-  I have updated the test names i.e. made the names more aligned with what user should expect by doing certain actions in the browserAction popup.
- Removed unnecessary checking (which are not aligned with the purpose of the test description) that might overwhelm a developer.

I have added comment on respective areas where I think some improvement was done.

src/lib/ext-popup.js
atiqueahmedziad: We have [enabled an Eslint rule to show error for console messages](https://github.com/mozilla/extension-activity-monitor/blob/master/.eslintrc.json#L9). Should we remove this?

